PR #916, opened 2022-05-26
Pick and place and image portrait hotfix develop

robhillman97:
Fixes the overwritten code discussed in todays meeting for pick and place and the Image augmentation


cawke (approved):
Thanks

PR #920, opened 2022-05-27
Fix for issue #919

### LINKED ISSUE #919: Pick&place and image marker augmentations missing
The pick and place and the image marker augmentations are missing in the list - although they are listed in the BrandConfiguration.cs under spareListOfAugmentations, which should be responsible for activating the available augmentations based on the platform. Possibly this is because 'pick&place' is used and not 'pickandplace' (the part of the variable name and content type in ContentType.cs)? Similar with 'image marker' which might have to be 'imagemarker' to work?

RC 1.9, 27.5.2022, iPad

jafarigit:
This will brings back the missing augmentation (image marker and pick and place) to the mobile version
Closes #919

### DIFF
--- a/Assets/MirageXR/Player/Scripts/Managers/BrandManager.cs
+++ b/Assets/MirageXR/Player/Scripts/Managers/BrandManager.cs
@@ -35,7 +35,7 @@ namespace MirageXR
 
 #if UNITY_ANDROID || UNITY_IOS
 
-    private readonly string[] spareListOfAugmentations = { "image", "video", "audio", "ghost", "label", "act", "vfx",  "model", "character", "pick&place", "image marker", "plugin" };
+    private readonly string[] spareListOfAugmentations = { "image", "video", "audio", "ghost", "label", "act", "vfx",  "model", "character", "pickandplace", "imagemarker", "plugin" };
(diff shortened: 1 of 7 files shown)


fwild (changes requested):
Can you please also fix the line 40 for hololens?

fwild:
Or was the hololens not affected?

fwild (approved):

Assets/MirageXR/Player/Scripts/Managers/BrandManager.cs
```diff
@@ -35,7 +35,7 @@ public class BrandManager : MonoBehaviour
 
 #if UNITY_ANDROID || UNITY_IOS
 
-    private readonly string[] spareListOfAugmentations = { "image", "video", "audio", "ghost", "label", "act", "vfx",  "model", "character", "pick&place", "image marker", "plugin" };
+    private readonly string[] spareListOfAugmentations = { "image", "video", "audio", "ghost", "label", "act", "vfx",  "model", "character", "pickandplace", "imagemarker", "plugin" };
```
cawke, line 38: is there any reason to keep two identical lists or can we keep just one? 
```
BrandManager.spareListOfAugmentations
BrandConfiguration.spareListOfAugmentations
```

Issue #877, opened 2022-04-07
Work through StyleCop messages


wekitecs:
In GitLab by @BenediktHensen on Apr 7, 2022, 15:58

Currently, StyleCop flags about 31000 warnings where code style is violated. We should go through it and fix them so that we get a consistent code style. This is also a prerequisite for meaningful automatic notifications of StyleCop as we only want to see them when new code violations or missing documentation is flagged.

PR #935, opened 2022-06-02
Fixed SA1005 warnings

BenediktHensen:
This pull request contributes to #877 and fixes the SA1005 warnings. For readability purposes, there should be a space in front of single-line comments.

The project still contains some of these warnings but they all relate to code which is commented out. It would be good to get rid of this at some point but for this, we need to decide what to do with the commented out code on a case-by-case basis.

It also fixes some typos in the comments along the way.


cawke (approved):
Thanks

PR #937, opened 2022-06-03
Character audio player fix

### LINKED ISSUE #917: Audio on character model broken
The audio recording on the character model augmentation does not work. It does record, but does not replay. The audio augmentation is created and can be played directly in the editor, by clicking on the audio augmentation in the list of augmentations, then clicking on play. Replay does not work, however, in playmode or editmode, and neither by directly clicking on the play button to the right of the head of the character model. I suspect something crashes connected to lip sync or it is simply the AudioClip.Play that is wrongly addressed?

(RC 1.9, ipad, 27.5.2022; likely to exist across all platforms)

dupes: https://github.com/WEKIT-ECS/MIRAGE-XR/issues/767

jafarigit:
Audio player of the characters works on mobile now
Closes #917


cawke (approved):
Thanks

Issue #795, opened 2022-01-20
Pick & place: sound effect for correct location?


wekitecs:
In GitLab by @wild on Jan 20, 2022, 15:19

There was a user proposal for a sound effect upon correctly placing to the target location - add?

wekitecs:
In GitLab by @robhillman97 on Jan 25, 2022, 11:50

What do we think about this? we now have the colour change, which helps with recognising a correct placement. I think this would add to that but I'm not sure its 100% necessary. I also think if we add a sound affect for correct placement we should add one for an incorrect placement as well, as currently the only indication that the placement wasn't correct is the arrow bouncing back.

wekitecs:
In GitLab by @wild on Feb 16, 2022, 18:51

Agree, but both sound effects would be good to have! How about these:
- correct: https://artlist.io/sfx/track/50543/magic-spells---chime-bells,-delight-win-ding
- wrong: https://artlist.io/sfx/track/77111/clear-ui---access-denied,-error

wekitecs:
In GitLab by @cawke on Mar 17, 2022, 03:18

need to check

PR #976, opened 2022-06-13
Pick and Place sound effects and improvemets

robhillman97:
Added sound effects to pick and place to address #795. 

Also the arrow object size is now saved with other pick and place data and bounding boxes have been added to the arrow and sphere


Assets/MirageXR/Common/Scripts/CombinedEditor/ActionDetailView/Annotation Editors/PickAndPlace/Pick.cs
```diff
@@ -91,6 +98,10 @@ void Update()
                 }
                 transform.hasChanged = false;
             }
+            else if(moveMode && pickOb.GetComponentInChildren<Renderer>().material.color != originalArrowColor)
```
cawke: cache this value - `pickOb.GetComponentInChildren<Renderer>()` and use it instead.

Assets/MirageXR/Common/Scripts/CombinedEditor/ActionDetailView/Annotation Editors/PickAndPlace/Pick.cs
```diff
@@ -178,26 +191,42 @@ public void ManipulationStart()
...
-            {
-                pickOb.transform.localPosition = new Vector3(resetPos.x, resetPos.y, resetPos.z);
+                {
+ 
+                    pickOb.transform.localPosition = new Vector3(placeLocation.localPosition.x, placeLocation.localPosition.y, placeLocation.localPosition.z);
 
+                    pickOb.GetComponentInChildren<Renderer>().material.SetColor("_Color", Color.green);
+                   
+                    if (shouldPlaySound)
+                        playAudio(correctAudio);
```
cawke, line 210: add brackets

Assets/MirageXR/Common/Scripts/CombinedEditor/ActionDetailView/Annotation Editors/PickAndPlace/Pick.cs
```diff
@@ -178,26 +191,42 @@ public void ManipulationStart()
...
 
-                pickOb.GetComponentInChildren<Renderer>().material.SetColor("_Color", originalArrowColor);
-            }
+                else if (resetOnMiss)
+                {                  
+                        pickOb.transform.localPosition = resetPos;
 
+                        pickOb.GetComponentInChildren<Renderer>().material.SetColor("_Color", originalArrowColor);
+                        if (shouldPlaySound)
+                            playAudio(incorrectAudio);
```
cawke: add brackets

cawke (changes requested):

Assets/MirageXR/Common/Scripts/CombinedEditor/ActionDetailView/Annotation Editors/PickAndPlace/Pick.cs
Alvareon: yeah, we should avoid is calling GetComponent every frame if we don't have to

Alvareon: yeah, we should avoid is calling GetComponent every frame if we don't have to

cawke (approved):
Thanks

PR #984, opened 2022-06-16
Change lrs dropdown into master for ARETE

robhillman97:
Merge lrs dropdown changes into master for ARETE pilot testing


robhillman97:
Converted to a draft as one of the references on the service bootstrapper is not being remembered after being pushed. Looking into it now

fwild (approved):

Issue #797, opened 2022-01-24
Possibility to delete account (Apple requirement)


wekitecs:
In GitLab by @fominykh on Jan 24, 2022, 14:43

By 30.06.2022, we need to add a feature to delete user account.
The new requirement by Apple is that if the app allows to create a user account, it must also allow to delete it.
Currently, we do not have account creation in the mobile app, but I think there is a plan to add this feature.

wekitecs:
In GitLab by @fominykh on Jan 24, 2022, 14:43

changed due date to June 01, 2022

fominykh:
In Moodle, I do not see a possibility to delete my user account from my profile. Reading forums, people say that there should be an option for a regular user to request account deletion, which triggers a message to an admin. This is a bit too manual... We need to investigate how to allow our Moodle users to delete their profiles with an option to delete all user data.
@cawke needs an API function to allow this in the app.

fwild:
Found this: https://moodle.org/plugins/local_gdpr_deleteuserdata (but it was built for Moodle 3.5, may not work anymore?)

fwild:
But I think this has moved into Moodle core - something along the lines of this: https://support.moodle.com/support/solutions/articles/80000946035-delete-my-data-from-a-moodlecloud-site 

I cannot find options to enable the data requests in Moodle, though - for me they do not show up under profile...?

fwild:
https://docs.moodle.org/36/en/Data_privacy

fwild:
and it says it is part of Moodle core since 3.5: https://moodle.org/plugins/tool_dataprivacy

fwild:
found it. needed to switch on: 

> Contact the privacy officer tool_dataprivacy | contactdataprotectionofficer
> Default: No
> If enabled, users will be able to contact the privacy officer and make a data request via a link on their profile page.

Now anyone can click on this link: https://learn.wekit-ecs.com/admin/tool/dataprivacy/createdatarequest.php?type=2
and request deletion.

fwild:
I propose we simply link to this?

fominykh:
So, there is still a manual step for an admin to process a deletion request, right? I guess it is sufficient for Apple.

fwild:
We can activate the 'confirm deletion requests automatically', but I think it is too dangerous - and users could easily delete themselves by mistake?

fominykh:
[...]

PR #990, opened 2022-06-23
Draft: Delete account button for #797

robhillman97:
Added Delete account button to mobile and HoloLens UI which opens the relevant Moodle page to address #797

delete account HoloLens
Delete account Mobile

Marked as draft as needs testing


Assets/MirageXR/Player/Scripts/Mobile/Popup/LoginView.cs
```diff
@@ -113,6 +115,10 @@ private void OnClickRegister()
     {
         Application.OpenURL(DBManager.registerPage);
     }
+    private void OnClickDeleteAccount()
+    {
+        Application.OpenURL(DBManager.deleteAccount);
```
cawke, line 120: It won't pass Apple's inspection. We need to delete the account from the app, not open the page in the browser. If we don't have access to the API, we can open the page in webview directly in the app. but we have to open the page with the user already logged in.

cawke:
I think we can close this for now

PR #992, opened 2022-06-29
Update sorting screen

Alvareon:
- updated toggles and text;
- moved PlayerPrefs to DBManager;
- fixed toggle transition;


cawke (approved):
TY

Issue #924, opened 2022-06-01
Calibration marker instruction: "button" instead of "close"


fwild:
The button on the laser eyes calibration hint dialogue has "button" instead of "close". This had been fixed in a previous version, but must have been reintroduced somehow?

iPad RC 1.9 (30.5.2022)

fwild:
Btw: this happens only on iPad, not on Android.

fwild:
Correction: this also happens on Android on 1.9.1 - must have overwritten an earlier correction in a preview release.

PR #997, opened 2022-07-14
924 marker close

Callustian:
Linked to #924

### DIFF
--- a/Assets/MirageXR/Player/Resources/Prefabs/Calibration/CalibrationGuide.prefab
+++ b/Assets/MirageXR/Player/Resources/Prefabs/Calibration/CalibrationGuide.prefab
@@ -78,7 +78,7 @@ MonoBehaviour:
     m_HorizontalOverflow: 1
     m_VerticalOverflow: 1
     m_LineSpacing: 1
-  m_Text: OK
+  m_Text: Close
 --- !u!1 &116997207649610675
 GameObject:
   m_ObjectHideFlags: 0
--- a/Assets/UnitTests/PlayModeTests/CalibrationTests.cs
+++ b/Assets/UnitTests/PlayModeTests/CalibrationTests.cs
@@ -43,7 +43,6 @@ namespace Tests
         GameObject personContainer;
         GameObject detectableContainer;
         GameObject sensorContainer;
-        GameObject guide;
 
         // dummy manager objects
         UiManager dummyUiManager;
@@ -173,11 +172,6 @@ namespace Tests
             detectableContainer = new GameObject("Detectables");
             sensorContainer = new GameObject("Sensors");
 
-#if UNITY_ANDROID || UNITY_IOS
-            guide = (GameObject)AssetDatabase.LoadMainAssetAtPath("Assets/MirageXR/Player/Resources/Prefabs/UI/Mobile/CalibrationGuide.prefab");
-#else
-            guide = (GameObject)AssetDatabase.LoadMainAssetAtPath("Assets/MirageXR/Player/Resources/Prefabs/Calibration/CalibrationGuide.prefab");
-#endif
             // register scene selection service
             referenceServiceConfiguration = new ActivitySelectionSceneReferenceServiceConfiguration
             {
@@ -695,26 +689,6 @@ namespace Tests
                 }
             }
         }
-
-        [UnityTest, Order(10)]
-        public IEnumerator Text_VerifyCloseText()
-        {
-            yield return EnsureTestReadiness();
-
-#if UNITY_ANDROID || UNITY_IOS
-            var ok = guide.transform.FindDeepChild("Button");
-            var component = ok.GetComponent<Button>();
-            var label = component.GetComponentInChildren<TMP_Text>();
-            var text = label.text;
-#else
-            var ok = guide.transform.FindDeepChild("OKButton");
-            var component = ok.GetComponent<Button>();
-            var componentText = component.GetComponentInChildren<Text>();
-            var text = componentText.text;
-#endif
-            Assert.AreEqual("Close", text);
-        }
-
 
         //[UnityTest, Order(99)]
         public IEnumerator PauseForDebug()
--- a/Assets/UnitTests/PlayModeTests/CalibrationTests.cs
+++ b/Assets/UnitTests/PlayModeTests/CalibrationTests.cs
@@ -173,8 +173,11 @@ namespace Tests
             detectableContainer = new GameObject("Detectables");
             sensorContainer = new GameObject("Sensors");
 
+#if UNITY_ANDROID || UNITY_IOS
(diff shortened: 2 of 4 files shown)


Assets/UnitTests/PlayModeTests/CalibrationTests.cs
```diff
@@ -687,6 +696,26 @@ public IEnumerator FocusOnPois_CheckLocalScales_z()
             }
         }
 
+        [UnityTest, Order(10)]
```
cawke: I don't think a test is needed to check what text is written on the button.

Assets/MirageXR/Player/Resources/Prefabs/Calibration/CalibrationGuide.prefab
```diff
@@ -78,7 +78,7 @@ MonoBehaviour:
     m_HorizontalOverflow: 1
     m_VerticalOverflow: 1
     m_LineSpacing: 1
-  m_Text: OK
```
cawke, line 81: I'm not sure if `ok` should be changed to `close`

cawke (changes requested):
I think it would just be enough to change the `button` text to `ok` in the `Assets/MirageXR/Player/Resources/Prefabs/UI/Mobile/CalibrationGuide.prefab`

Assets/MirageXR/Player/Resources/Prefabs/Calibration/CalibrationGuide.prefab
Callustian: Hi @cawke This is a regression problem. At some point this text was added and then, for some reason, returned to the 'OK' value. (I understand the test for 'OK' is 'Close' for IOS, Android and Windows). So using TDD principles, I added a test for this.

Assets/UnitTests/PlayModeTests/CalibrationTests.cs
Callustian: Hi @cawke  This is a regression problem. At some point this text was added and then, for some reason, returned to the 'OK' value. (I understand the test for 'OK' is 'Close' for IOS, Android and Windows). So using TDD principles, I added a test for this. Is this argument convincing to you? Thanks.

cawke: I understand, but the test is too closely related to the things that it should not check. For example, if the prefab will be moved to another folder, or in the prefab will change the name of the button or added another button with the same name, these changes will not pass the test. I believe that this test will only bring more inconvenience than any use.

Callustian: Honestly, I think these are arguments to keep the test, instead of deleting the test. If you make a change in the software and you don't change the set of tests, a test should detect the change. Tests are there to detect changes that could inadvertently break the software. I think, though, that the place I wrote the test in is wrong. Would we find an agreement in relocating the test to a suite of its own? Thanks.

cawke: I still think that this test is only an inconvenience, no matter where it is located. We should cover more complicated things with tests, not like button text.

@Alvareon and @BenediktHensen please give us your opinion.

Alvareon: Well, in my opinion, we shouldn't cover texts by tests. Our project has too many assets with texts. It doesn't really matter. We need to focus on the functionality.

Callustian: @Alvareon . I submit regressions cost time, and once you have learned how to write a test for one item of text, it is immediate to apply that knowledge for a new element. In addition, if I may, this documents the structure of our project. Thanks.

cawke: My complaints about the test are mainly about how you find the prefab (direct path to the file) and how you find the component (by name throughout the prefab)

image

image

Callustian: @cawke I understand that the path is relative and does not change, and so it is about the tag of the component. I could think of other ways of retrieving the prefab (... suggestions?). We don't have a table for the tags used for the elements of the prefab, I submit adding this table is the solution to make this test more architecture agnostic.

BenediktHensen: I would also say that we should focus on testing the functionality. Ideally, there should be a separation between the functional layer and the visuals of the scene. So, the methods that we test are separated from the actual GameObjects, scene structure, prefabs, etc. This way, the method will work with just a minimal setup and there is no need to replicate the scene structure by loading prefabs.

The main trouble with such tests that require visual elements is the fact that we need to generate these visual scenes from the test's code. Here, retrieving assets from paths both on disk but also in the Unity scene can break the test easily. The same is true for some of the existing tests which start re-creating scene structure in the initialization code. If we move around some files, rename folders or GameObjects, this will break the test, although the functionality that the test checks would still work.

Callustian: @BenediktHensen  I understand your concerns, and they are reasonable. A test is meant to detect a problem, though, not necessarily to last for years. The solution, I think, is to use a Localisation table and verify the constant we use for the component. This way we decouple the component and the text, and we verify the text only. There is still the issue of some person inadvertently changing the reference in the component and I am unsure if the team wants to introduce localisation tables at this point, though.

Callustian: A test that works is more than nothing.

fwild: Final though (I see this is already resolved): such test could be useful in the context of localisation - when we have language support for different languages (to check whether all or at least key items are actually updated when the language is switched)?

cawke (changes requested):
please remove the changes in the files `Assets/UnitTests/PlayModeTests/CalibrationTests.cs` and `Assets/UnitTests/PlayModeTests/PlayModeTests.asmdef`

Callustian:
@cawke I deleted the test in `Assets/UnitTests/PlayModeTests/CalibrationTests.cs`. I didn't add changes to `Assets/UnitTests/PlayModeTests/PlayModeTests.asmdef`

cawke:
@Callustian you deleted only the test, but you left other changes in the file.
https://github.com/WEKIT-ECS/MIRAGE-XR/pull/997/files

image
image

Callustian:
Hello @cawke I only changed `Assets/MirageXR/Player/Resources/Prefabs/Calibration/CalibrationGuide.prefab`  that is a requested change and `Assets/UnitTests/PlayModeTests/CalibrationTests.cs`. Unity, automatically made all the other changes. If you don't want to add them, I think is more immediate for me to close this pull request and make some other branch with only the change in the prefab.

PR #1018, opened 2022-08-11
Further stylecop fixes

BenediktHensen:
This pull request fixes various stylecop warnings and adjusts the ruleset file so that it only contains rules which make sense for the project.
Contributes to #877.

Fixed warnings include:
- SA1000: put a space after the if keyword
- SA1001: put a space after a comma
- SA1002: put a space after a semicolon
- SA1013: no space in front of a closing brace
- SA1028: removed whitespace at the end of lines
- SA1106: removed successive semicolons which created empty statements
- SA1116: organization of parameters for long method calls
- SA1117: organization of parameters for long method headers
- SA1309: no underscore in front of local variables
- SA1626: only two slashes for comments which are not XML documentation
- SA1616: documented return values
- SA1622: documetend generic wildcard types


Assets/MirageXR/Common/Scripts/ActivityCard/ActivityCard.cs
```diff
@@ -10,28 +10,28 @@ public class ActivityCard : MonoBehaviour
     {
         // Define all the components.
 
-        private AudioSource _audio;
+        private AudioSource audio;
```
cawke: I'm not sure we should change the `_privateField` style to `privateField`. Maybe we should take a vote? @Alvareon @BenediktHensen @Callustian @jafarigit @robhillman97

Alvareon: I think it is a matter for mutual arrangement. @BenediktHensen, what problem do we want to solve? What's wrong with old name?

BenediktHensen: Yes, it is just a matter of taste. I went with the default rule settings of Stylecop but if we want, we can vote on the style that we want to use.

The important part is that it should be consistent in the project so that the code looks more homogeneous. Currently, we have all three styles for private fields in the project: no underscore "variableName", an underscore in the beginning "_variableName" and "m_" at the beginning "m_variableName".

cawke: Let's just vote. 
I vote for the option `_privateField`
@Alvareon @BenediktHensen @Callustian @jafarigit @robhillman97

Alvareon: I vote for  `_privateField`

fwild:
Can we integrate this merge request? Or is there still need to act?

fwild:
@BenediktHensen can we merge this or is there still need to change anything (or has it been superseded by your other changes?).

PR #1027, opened 2022-08-24
UI Kit

Alvareon:
1. Created the set of basic prefabs:
- Texts;
- Buttons;
- TextFields;
- Toggles;

2. Updated screens in accordance with the UI Kit.
3. Added new font SFProDisplay-Regular.


cawke (approved):
TY

PR #1057, opened 2022-10-06
Update windows_workflow.yml

fwild:
changed old unity version to new one: 2020.3.38f1


fwild:
breaks, fetches wrong unity image (base-0 instead of base-1.0.1)

cawke (approved):
if I understand correctly, then the build for windows completed successfully. TY

PR #1060, opened 2022-10-06
Update linux_workflow.yml

fwild:
Updated to match Unity version (2020.3.38f1) - changed twice here (forgot first time around)


fwild:
breaks, fetches wrong unity image :(

PR #1080, opened 2022-10-25
character triggers added

robhillman97:
Character AI can now recognise "%%trigger%%" commands to move forward a step.


Assets/MirageXR/Player/Scripts/IBMWatson/DialogueService.cs
```diff
@@ -150,10 +149,12 @@ public void SendMessageToAssistant(string theText)
 
     }
 
-    private void MakeDance()
+
+    private void nextStep()
```
cawke:
```suggestion
    private void NextStep()
```

cawke (changes requested):

cawke (approved):
ty

PR #1091, opened 2022-11-01
Update linux_workflow.yml

fwild:
added new html reports interface (as old one does not support unity nunit)


fwild:
not sure its worth doing - any opinions about this additional report format?

PR #1092, opened 2022-11-01
Update linux_workflow.yml

fwild:
convert existing nunit to junit reports so that dorny html reports work


fwild:
don't merge - path to -report.xml files is not correct

PR #1093, opened 2022-11-01
Update linux_workflow.yml

fwild:
path fixes again for html reports (dorny)


fwild:
don't merge - path to -report.xml files is not correct

PR #1149, opened 2022-12-15
Hotfixes rob

### LINKED ISSUE #816: Marker augmentation: display "find this" + marker image
In Play Mode: When the marker augmentation is not yet identified (tracking not yet found, position of the task station not yet set to new world anchor, subsequent tracking stopped?), there should be a screen-space dialogue that instructs the user to find the marker, also displaying the marker. 

In this context: user reports: "The image marker would be more useful if there was an option to display the
educational scenario only when the specific marker would be spotted."

UPDATE: when this functionality moves into the task station, it still makes sense - and it is already foreseen in the new UI!

### LINKED ISSUE #971: Add preview activity
Screenshot 2022-06-16 at 15 20 04

Screenshot 2022-06-16 at 15 20 42

### LINKED ISSUE #1045: QR code from Moodle opens app, but does not trigger download
We have this feature in Moodle that you can get a link to the activity to start it. It seems it has stopped working for me. Pointing the native OS camera app at the code gives me the link, and the link opens the app (as it is correctly registered for the new protocol "wekit://"), but the app does not initiate the download of the activity. 

QR-code

I have tested: with app not started (starts the app, but no download trigger); with app already started (doesn't download or open); with app already started and activity already previously downloaded (activity is not started).

### LINKED ISSUE #1139: Trigger for actions broken
If a trigger is set for an 'action' augmentation, it does not save, and is not executed in play mode.

### LINKED ISSUE #1142: When switching from 'edit step' to playmode, the edit dialogue stays open
When switching from 'edit step' to playmode, the edit dialogue stays open, and e.g. the delete function is still accessible. 

Workaround: Navigate back to the action step list (by pressing "Back to activity" top left). 

Solution: close the edit dialogue and go back to activity automatically.

### LINKED ISSUE #1145: update text in calibration video dialogue
The calibration video dialogue is still useful - it is a good reminder that you need to calibrate!
But the calibration has now moved into the submenu, so the text on the dialogue should
say:

"Please run the calibration routine first by tapping on the 'Calibration' tab visible once you close this dialogue. If you get stuck, don't worry: The help menu can explain in more detail the why and how."

### LINKED ISSUE #1146: Add jump to option for audio trigger
Currently, in RC 2, the audio augmentation has only a trigger next step option, no option to jump to a specific step - the trigger behaviour should be consistently the same across all 'triggerable' augmentations.

robhillman97:
If a step has an image marker a preview button will appear on the action step list which will show the marker image, closes #816 

Preview button now sets play mode and opens 1st action, closes #971 

Warning given when trying to download activity from the wrong moodle repo via QR code, closes #1045 

fix for issues with action trigger, closes #1139 

switching to play mode will open the action list, closes #1142 

updated calibration video dialougue , closes #1145 

You can now pick the step an audio trigger jumps to, closes #1146


cawke (changes requested):
can you run tests localy?

image

Assets/MirageXR/Player/Scripts/Augmentations/AudioPlayer/AudioPlayer.cs
```diff
@@ -200,7 +200,7 @@ private static IEnumerator ActivateTrigger(AudioSource audioSource, Trigger trig
                         activityManager.ActiveAction.isCompleted = true;
                     }
 
-                    activityManager.ActivateNextAction();
+                    activityManager.ActivateActionByIndex(int.Parse(trigger.value)-1);
```
cawke: add check

cawke (changes requested):

robhillman97:
UnitTests
UnitTests2

cawke (approved):

PR #1163, opened 2022-12-16
Sorting view

### LINKED ISSUE #1154: Sorting view: filter/sort by does not work
Filters and sort by are not working. The activity list always displays "all", no matter whether I select "show all", "my assignments", or "my activities"). 

Sort by also does not have an effect - no matter whether I select "by relevance" or "by date", it always displays "by date".

The sorting view also forgets my selection when I return to it. 

The card view size selection works!

robhillman97:
Adds sorting view functionality.

closes #1154

### DIFF
--- a/Assets/MirageXR/Player/Scripts/Mobile/PopupsViewer.cs
+++ b/Assets/MirageXR/Player/Scripts/Mobile/PopupsViewer.cs
@@ -92,10 +92,14 @@ public class PopupsViewer : MonoBehaviour
 
     private void OnClose(PopupBase popup)
     {
-        if (_stack.Peek() == popup)
+        if (_stack.Count > 0)
         {
-            Destroy(_stack.Pop().gameObject);
-            UpdateView();
+
+            if (_stack.Peek() == popup)
+            {
+                Destroy(_stack.Pop().gameObject);
+                UpdateView();
+            }
         }
     }
 }
--- a/Assets/MirageXR/Player/Scripts/Mobile/PopupsViewer.cs
+++ b/Assets/MirageXR/Player/Scripts/Mobile/PopupsViewer.cs
@@ -92,14 +92,16 @@ public class PopupsViewer : MonoBehaviour
 
     private void OnClose(PopupBase popup)
     {
-        if (_stack.Count > 0)
+        if (_stack.Count <= 0)
         {
+            Debug.LogError("Stack is empty!");
+            return;
+        }
 
-            if (_stack.Peek() == popup)
-            {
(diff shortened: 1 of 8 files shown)


Assets/MirageXR/Player/Scripts/Mobile/PopupsViewer.cs
```diff
@@ -92,10 +92,14 @@ private void OnOutTap()
 
     private void OnClose(PopupBase popup)
     {
-        if (_stack.Peek() == popup)
+        if (_stack.Count > 0)
```
cawke: If you get an error from PopupViewer that the stack is empty, then you have done something wrong. 
which is how you use toggle when you change the toggle, the `ShowSmallCard `and `ShowBigCard `method will be called, which will cause the PopupViwer to close twice. This also causes ActivityListView to be updated twice, which takes a very long time. Ideally, we don't need to pull data from the server when we change the card display. We should just use the data we already have.

image

But a check can be added.
```
    private void OnClose(PopupBase popup)
    {
        if (_stack.Count <= 0)
        {
            Debug.LogError("Stack is empty!");
            return;
        }

        if (_stack.Peek() == popup)
        {
            Destroy(_stack.Pop().gameObject);
            UpdateView();
        }
    }
```

Assets/MirageXR/Player/Scripts/Moodle/DBManager.cs
```diff
@@ -154,6 +162,23 @@ public static LearningRecordStores publicCurrentLearningRecordStore
...
+        {
+            get => CurrentSortby;
+            set => CurrentSortby = value;
+        }
+
+        public static string Username
+        {
+            get => username;
+        }
+
```
cawke: what's the point?

cawke (changes requested):

Assets/MirageXR/Player/Scripts/Mobile/PopupsViewer.cs
robhillman97: Ah I see the issue, just pushed a fix. Though these toggles call UpdateView in ActivityListView_v2 which destroys the existing ActivityListItems and remakes them using the list of session containers we already have, so isn't pulling data from the server every time we change the card size. Unless I am mistaken.

Assets/MirageXR/Player/Scripts/Moodle/DBManager.cs
robhillman97: My mistake, made this variable thinking I'd need it then didn't use it. Has been removed

cawke (approved):

Issue #1224, opened 2023-01-18
Update UI v2 according to the design comments in Figma


Alvareon:
The design comments: 
https://www.figma.com/file/AqAGvX0KRDNRlVz3KFzoMH/New-Concept?node-id=2213%3A19785&t=EEG49AhdBpvPmhOl-0

Alvareon:
https://github.com/WEKIT-ECS/MIRAGE-XR/pull/1225 - part1

Alvareon:
https://github.com/WEKIT-ECS/MIRAGE-XR/pull/1231 - part 2

Alvareon:
https://github.com/WEKIT-ECS/MIRAGE-XR/pull/1236 - part 3

Alvareon:
https://github.com/WEKIT-ECS/MIRAGE-XR/pull/1239 - part 4

fwild:
few items still missing, new issues created / coming

PR #1231, opened 2023-01-23
Design v2 update part2

Alvareon:
Update UI v2 according to the design comments in Figma:    https://github.com/WEKIT-ECS/MIRAGE-XR/issues/1224


cawke (approved):
ty

PR #1257, opened 2023-02-10
Illegal character zip check

### LINKED ISSUE #1222: Check for Illegal characters when Unzipping ARLEM file
The "William's Family activity" contains an 3D model called; Whiskey "White Horse". This prevents the activity from being downloaded correctly as the app tries to create a new folder for the model with the same name throwing an "Illegal characters in path" error due to the " symbol not being valid in folder names. 

In ZipUtilities.cs the method ExtractZipFileAsync should check the entryFileName Variable for illegal characters and remove them.

The illegal characters are  \ / : * ? " < > |

robhillman97:
Added CheckFileForIllegalCharacters method to zip utilities which checks a file name or path for illegal file characters and removes them. The new method is used when zipping and unzipping an activities ARLEM file and when creating a 3D model augmentation.

Closes #1222

### DIFF
--- a/Assets/MirageXR/Common/Scripts/ZipUtilities.cs
+++ b/Assets/MirageXR/Common/Scripts/ZipUtilities.cs
@@ -2,7 +2,6 @@ using System;
 using System.IO;
 using System.Threading.Tasks;
 using ICSharpCode.SharpZipLib.Zip;
-using UnityEngine;
 
 /// <summary>
 /// Utilities for handling ZIP files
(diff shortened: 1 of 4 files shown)


Assets/MirageXR/Common/Scripts/ZipUtilities.cs
```diff
@@ -119,4 +121,24 @@ public static async Task AddFileToZipStreamAsync(ZipOutputStream zipStream, stri
...
+        foreach (char c in characters.ToCharArray())
+        {
+            if (file.Contains(c))
+            {
+                file = file.Replace(c.ToString(), string.Empty);
+            }
+        }
+
+        return file;
+    }
```
cawke:
```suggestion
    public static string CheckFileForIllegalCharacters(string fileName)
    {
        const string characters = "[\\:*?\"<>|]";
        return Regex.Replace(fileName, characters, string.Empty);
    }
```

cawke (changes requested):

cawke (approved):

PR #1258, opened 2023-02-10
Persisted Logout

### LINKED ISSUE #1214: logout not persisted
When you log out and quit the app, next time started you will be logged in again. During a session it, however, correctly logs out. Just not persisted (= logout does not seem to clear stored username/password).

robhillman97:
username and password are now cleared from config.info when a user logs out

Closes #1214


cawke (approved):
ty

PR #1259, opened 2023-02-10
Sharing the calibration image

Alvareon:
Added function to share the calibration image to email, messengers or printer directly from the application. 
The user can quickly print the calibration image. For convenience, the image file is converted to pdf format of A4 size. 

demo_sharing_resize

The file is stored in Unity Binary Text Asset Data format in the application.


cawke (approved):
ty

Issue #1217, opened 2023-01-17
autosave activity title update


fwild:
Updates to the activity title should (like everything else) autosave, so that the activity list is updated when going back to "home". (At the moment, the list only updates, if one explicitly goes to "publish..." And hits "local"+"save").

PR #1266, opened 2023-02-15
Auto save update

### LINKED ISSUE #1234: Auto-safe for keep alive
How to reproduce:
1. In step 1, add an image augmentation.
2. Create step 2.
3. Open step 1, select the earlier created image augmentation, and select keep alive 1-2.
4. Open step 2, and observe that the same image augmentation also appears in step 2.
5. [updated] Autosaving does not work here. Close the app on this step without saving in the "Publish" menu.
6. Open activity. And observe that step 2 does not exist anymore.

The bug does not appear if I "save-to-local-device" on step 5.

Related to: https://github.com/WEKIT-ECS/MIRAGE-XR/issues/1217

robhillman97:
Changing the activity title/Description, an activity title/description and adding/changing keep alive on an augmentation now saves the activity locally 

closes #1234 and #1217

### DIFF
--- a/Assets/MirageXR/Tests/NewUI/ContentListView_v2.cs
+++ b/Assets/MirageXR/Tests/NewUI/ContentListView_v2.cs
@@ -216,10 +216,7 @@ public class ContentListView_v2 : BaseView
         _currentStep.instruction.title = newTitle;
         EventManager.NotifyActionModified(_currentStep);
 
-        if (_info.activeInHierarchy)
-        {
-            activityManager.SaveData();
-        }
+        activityManager.SaveData();
     }
 
     private void OnStepDescriptionChanged(string newDescription)
@@ -227,10 +224,7 @@ public class ContentListView_v2 : BaseView
         _currentStep.instruction.description = newDescription;
         EventManager.NotifyActionModified(_currentStep);
 
-        if (_info.activeInHierarchy)
-        {
-            activityManager.SaveData();
-        }
+        activityManager.SaveData();
     }
 
     private void OnAnnotationSelected(ToggleObject annotation)
--- a/Assets/MirageXR/Tests/NewUI/ContentListView_v2.cs
+++ b/Assets/MirageXR/Tests/NewUI/ContentListView_v2.cs
@@ -215,12 +215,22 @@ public class ContentListView_v2 : BaseView
     {
         _currentStep.instruction.title = newTitle;
         EventManager.NotifyActionModified(_currentStep);
+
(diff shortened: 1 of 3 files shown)


Assets/MirageXR/Tests/NewUI/StepsListView_v2.cs
```diff
@@ -54,8 +54,8 @@ public class StepsListView_v2 : BaseView
     public override void Initialization(BaseView parentView)
     {
         base.Initialization(parentView);
-        _inputFieldActivityName.onValueChanged.AddListener(OnActivityNameChanged);
-        _inputFieldActivityDescription.onValueChanged.AddListener(OnActivityDescriptionChanged);
+        _inputFieldActivityName.onEndEdit.AddListener(OnActivityNameChanged);
+        _inputFieldActivityDescription.onEndEdit.AddListener(OnActivityDescriptionChanged);
```
cawke: if you changed event, pleace change `OnActivityNameChanged` the name to `OnActivityNameEndEdit` too.
same with OnActivityDescriptionChanged.

Assets/MirageXR/Tests/NewUI/ContentListView_v2.cs
```diff
@@ -215,12 +215,22 @@ private void OnStepNameChanged(string newTitle)
     {
         _currentStep.instruction.title = newTitle;
         EventManager.NotifyActionModified(_currentStep);
+
+        if (_info.activeInHierarchy)
+        {
+            activityManager.SaveData();
+        }
```
cawke: could you explain why there is a check?

cawke (changes requested):

Assets/MirageXR/Tests/NewUI/ContentListView_v2.cs
robhillman97: These checks prevent the activity being saved as soon as it is created as we set the names and descriptions when creating the activity. With the checks the user has to go to the correct page and change the name/description themselves for the activity to save

cawke: I understand, but I think this was relevant to the `onValueChanged` event. Now the `onEndEdit` event is used. But even if `onEndEdit` event is called when text is changed from code, I don't like this check because it doesn't guarantee unforeseen saving. It is better in such a case to unsubscribe the text fields from events, apply the new text and sign the text fields again.

cawke (approved):

PR #1271, opened 2023-02-16
Lighting fix + ambient color

cawke:
one light source and ambient color, instead of 4 light sources.


cawke:
Needs to be tested on Hololens

fwild (approved):

Issue #980, opened 2022-06-14
Mobile: Re-editing 3D model opens glyph edit panel


fwild:
When you tap on an existing 3D model augmentation in the action step's augmentation list, the 'action' glyph panel opens instead.

PR #1278, opened 2023-02-17
Incorrect content type for 3D models fix

### LINKED ISSUE #1204: 3d model augmentation: edit opens 'action' aug edit panel
When I add a 3D model, and then later click via context menu on 'edit', it brings up the wrong edit panel (that of 'action').
Dupe: https://github.com/WEKIT-ECS/MIRAGE-XR/issues/980

robhillman97:
Changed .contains to .StartsWith in ParsePredicate method of ContentType.cs as any 3d model which contained "act" in its name was considered an action glyph. This would also cause any 3d model with a name which contains "image", "video", "audio", "sound", "ghosttracks", "ghost", "label", or "vfx" to have the wrong content type. 

closes #1204 and #980


cawke (approved):
ty

PR #1281, opened 2023-02-21
Added WEKIT and ARETE privacy policy links and added open privacy policy button to profile page

### LINKED ISSUE #1152: Display privacy policy
Privacy policy / Terms of Use should be displayed the first time you start using the app and the user should agree to them.

robhillman97:
Closes #1152 

Added open privacy policy button to profile view and added privacy policy links for WEKIT and ARETE to DBManager. Open privacy Policy Button will open the correct link for the connected server

"Open Privacy Policy" button (Profile prefab changes):
Screenshot_PrivacyPolicy

### DIFF
--- a/Assets/MirageXR/Player/Scripts/Moodle/DBManager.cs
+++ b/Assets/MirageXR/Player/Scripts/Moodle/DBManager.cs
@@ -77,6 +77,10 @@ namespace MirageXR
         public const string WEKIT_URL = "https://learn.wekit-ecs.com";
         public const string ARETE_URL = "https://arete.ucd.ie";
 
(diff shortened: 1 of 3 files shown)


cawke:
what happens if the user enters his server address ("Other" button)? what page will it open when clicking on the "Open Privacy Policy" button?

robhillman97:
> what happens if the user enters his server address ("Other" button)? what page will it open when clicking on the "Open Privacy Policy" button?

Currently it will just open the privacy policy of either WEKIT or ARETE depending which was last used before entering their own server address. Agreed this should change but I am not sure if we need to allow them to add their own privacy policy link or just remove the button in this situation as it would be their server so a privacy policy is probably not necessary.

cawke (approved):

PR #1285, opened 2023-02-24
Changed Author unknown to you

### LINKED ISSUE #1219: display username as authorname for (new) local activities
Local activities that are not saved to the cloud currently list "author: unknown", but should probably say "you" or "$username" or something like that.

robhillman97:
Changed Author unknown to you in sessionContainer.cs

Closes #1219


cawke (approved):
I don't think it's a good idea, but let it stay that way for now.

PR #1344, opened 2023-04-17
brandConfiguration improvment

### LINKED ISSUE #1292: Argumentations array in BrandManager and BrandConfiguration
The argumentations array is modified in BrandManager.cs, and it should be same in BrandConfiguration.cs as well. "vfx" is replaced  by "effect" in BrandManagere, so it should be changed in  BrandConfiguration.cs too.

jafarigit:
brandConfiguration will use the augmentations list from BrandManager now, and no need to two arrays. Also some methods are refactored and the code of both classes are improved

UPDATE:
The Moodle URL and xAPI URL have been added to brandConfiguration, but they are not yet being used in the app. Developers who have refactored MoodleManager, DBManager, and other components can utilize these variables through the new static class ConfigParser by accessing ConfigParser.MoodleUrl and ConfigParser.XApiUrl.
BrandManager has also been refactored. The augmentations list will now be created using the ContentType enum, eliminating the need for two separate arrays. This change streamlines the process, getting rid of unnecessary splitting and replacing between augmentation names, predicates, and so on.

Closes #1292


cawke (approved):
ty

jafarigit:
The Moodle URL and xAPI URL have been added to brandConfiguration, but they are not yet being used in the app. Developers who have refactored MoodleManager, DBManager, and other components can utilize these variables through the new static class ConfigParser by accessing ConfigParser.MoodleUrl and ConfigParser.XApiUrl.

BrandManager has also been refactored. The augmentations list will now be created using the ContentType enum, eliminating the need for two separate arrays. This change streamlines the process, getting rid of unnecessary splitting and replacing between augmentation names, predicates, and so on.

@fwild, please ask the relevant developer(s) to integrate the ConfigParser data with the necessary UI and models. I attempted to use them in DBManager, but it seems that many things have changed since I last worked on it, and the URLs have become intertwined with various code segments.

fwild (approved):

Issue #1295, opened 2023-03-09
The calibration marker orientation should not affect the rotation of augmentations when they spawn


fominykh:
Currently,
1. Put the calibration marker vertically on a wall.
2. Start the app
3. Create new activity
4. Calibrate
5. Add Action glyph "Green Checkmark".
6. Observe that the glyph appears lying horizontally.
7. Move the calibration marker horizontally on the floor
8. Calibrate
9. Add Action glyph "Green Checkmark".
10. Observe that the glyph appears standing vertically.

Should be,
The green checkmark glyph should appear standing up vertically regardless of the position of the calibration marker.

This issue applies also to som other augmentation types.

fwild:
Orientation for calibration should always be: centre of the calibration marker + Vector3.Up

PR #1349, opened 2023-04-20
The calibration marker orientation should not affect the rotation of augmentations when they spawn

cawke:
Calibration has been optimized
The calibration marker orientation should not affect the rotation of augmentations when they spawn

https://github.com/WEKIT-ECS/MIRAGE-XR/issues/1295


fwild (approved):
looks correct - now needs testing!

PR #1350, opened 2023-04-20
Create new search screen

Alvareon:
Result:

search


cawke (changes requested):
remove space 

image

Alvareon:
done

cawke (changes requested):
The scroll view is larger than the view and the bottom elements are not visible

image

Alvareon:
> The scroll view is larger than the view and the bottom elements are not visible
> 
> image

fixed

cawke (approved):

PR #1353, opened 2023-04-20
Viewing tutorial activity changed wav file to mp3 and pick & place fix

robhillman97:
The ghost audio is now an mp3 to reduce file size. A small change was made to the audio player script to allow for mp3 files.

also remade the pick and place augmentation in this activity using 2.1 so should work fine now on mobiles


Assets/MirageXR/Player/Scripts/Augmentations/AudioPlayer/AudioPlayer.cs
```diff
@@ -389,7 +387,19 @@ private IEnumerator LoadAudio()
...
+                    audioClip = www.GetAudioClip(false, false, AudioType.WAV);
+                }
+                else if (filename.EndsWith("mp3"))
+                {
+                    audioClip = www.GetAudioClip(false, false, AudioType.MPEG);
+                }
+                else
+                {
+                    audioClip = www.GetAudioClip(false, false, AudioType.UNKNOWN);
+                }
```
cawke: move getting the type to a separate method.

and what happens if you try to load a file with type `AudioType.UNKNOWN`?
if it throws an error it needs to be handled.

```suggestion
                var audioType = GetAudioType(filename);
                AudioClip audioClip = www.GetAudioClip(false, false, audioType);
                
            private AudioType GetAudioType(string filename)
            {
                if (filename.EndsWith("wav"))
                {
                    return AudioType.WAV;
                }
                else if (filename.EndsWith("mp3"))
                {
                    return AudioType.MPEG;
                }
                else
                {
                    return AudioType.UNKNOWN;
                }
            }
```

cawke (changes requested):

cawke (approved):
ty

Issue #1338, opened 2023-04-10
Remove the Search button from the bottom panel on all screens except the Home/Activity screen


fominykh:

Alvareon:
https://github.com/WEKIT-ECS/MIRAGE-XR/pull/1363

fwild:
I think this is done everywhere - cannot find any other appearance?

fominykh:
yes, done https://github.com/WEKIT-ECS/MIRAGE-XR/pull/1363

PR #1363, opened 2023-04-24
Hide the search button when we leave the home screen

Alvareon:
https://github.com/WEKIT-ECS/MIRAGE-XR/issues/1338


fwild:
@Alvareon question: is the search icon still visible when you select an activity to view? instead of creating an activity?

Alvareon:
> @Alvareon question: is the search icon still visible when you select an activity to view? instead of creating an activity?

@fwild the search icon is only available from the home screen. The icon is not visible when we select an activity to view/edit:

search_button

cawke (approved):

PR #1365, opened 2023-04-27
arfoundation calibration fix

cawke:


fwild (approved):

cawke:
I changed the tracker behavior, which saves us from situations when ARFounadation recreates the ImageTarget, which causes us to lose links.

Issue #1378, opened 2023-05-01
After the image marker is detected, the spawning position does not move to it together with the diamond and the red sphere


fominykh:
Currently:
When an image marker is added to a step, the user can find the marker with the camera. After the marker is detected the diamond and the red spawning sphere move to the marker location. However, new augmentations spawn in the old diamond position.
Should be:
The new augmentation should spawn at the marker position.

PR #1382, opened 2023-05-02
spawning position fix

cawke:
https://github.com/WEKIT-ECS/MIRAGE-XR/issues/1378


Alvareon:
it seems that in the future it is necessary to reduce our technical debt )

Alvareon (approved):

Issue #1445, opened 2023-06-01
grid visualization


cawke:
https://github.com/WEKIT-ECS/MIRAGE-XR/issues/1327

Issue #1403, opened 2023-05-10
Snap to Grid: GridManager - settings for the grid.


cawke:
https://github.com/WEKIT-ECS/MIRAGE-XR/issues/1327

- allow to set grid size globally in profile > group: preferences
- allow to set rotation angles globally in profile > group: preferences
- standard size: 5 cm
- standard grid angle: 0 degrees

fwild:
set standard size for angles to 15 degrees.

PR #1446, opened 2023-06-01
Grid Manager

cawke:
based on https://github.com/WEKIT-ECS/MIRAGE-XR/pull/1442

https://github.com/WEKIT-ECS/MIRAGE-XR/issues/1445
https://github.com/WEKIT-ECS/MIRAGE-XR/issues/1403


Assets/MirageXR/Common/Scripts/Managers/PointCloudManager.cs
```diff
@@ -12,6 +12,10 @@ public class PointCloudManager : MonoBehaviour
 
     public async Task<bool> InitializationAsync()
     {
+#if !UNITY_ANDROID && !UNITY_IOS
+        return;
+#endif
+
```
robhillman97, line 18: return; throws and error as the method is expecting a bool. Change to return false;

robhillman97:
Merged into HololensFix branch, all works fine in editor with a couple of conflicts. The floor detection pop up gets in the way of the hand pointer so the interaction changes to near interaction (see how the interaction circle moves to the tip of my finger [here](https://drive.google.com/file/d/15SQ6ee_JM-G4gc4dtfr3oGAu_rLG8cvc/view?usp=sharing)). This can be resolved by either just moving the window up and back or by changing it so that it does not follow the user. You can also test this in editor by holding the space bar and using the simulated hands: 

Pointer pos

I have made a quick fix for this locally, just re-building to see if all works as expected

robhillman97:
There is an issue on HoloLens builds which prevents selecting the floor as we do not have the spatial map mesh active and visable. I have added a method to FloorManagerWrapper.cs which turns the spatial awareness mesh on and off when starting the detect floor routine:

    private void MeshOn(bool on)
    {
        var observer = CoreServices.GetSpatialAwarenessSystemDataProvider<IMixedRealitySpatialAwarenessMeshObserver>();

        if (on)
        {
            observer.DisplayOption = SpatialAwarenessMeshDisplayOptions.Visible;
        }
        else
        {
            observer.DisplayOption = SpatialAwarenessMeshDisplayOptions.None;
        }
    }

[However I am getting a NRE when selecting a mesh](https://drive.google.com/file/d/1vBuXCip4xu_p-Xye0bkhS5JMTrdX1EhS/view?usp=sharing). I think It is best to merge this branch into develop once the return error and pop up position have been fixed. I can then pull the updates from develop into the HoloLens fix branch, fix the conflicts and then we can work on the HololensFix branch to get the map working

fwild (approved):

PR #1508, opened 2023-07-13
pick and place fix

cawke:


fwild (changes requested):
please add that when user placed in correct position in view mode, and then switches to edit mode, the rest position is restored again in edit mode

Assets/MirageXR/Common/Scripts/CombinedEditor/ActionDetailView/Annotation Editors/PickAndPlace/PickAndPlaceController.cs
```diff
@@ -104,51 +65,68 @@ public override bool Init(ToggleObject obj)
...
 
-        public void SavePositions()
+        private void SavePositions()
         {
             if (_myObj == null || _myObj.poi == string.Empty || gameObject == null)
             {
                 return; // only if the poi is instantiated not the prefab
             }
+
+            if (!RootObject.Instance.activityManager.EditModeActive)
```
robhillman97: This check should be removed as it prevents changes being saved when in play mode. This means if you move the target or arrow in edit mode, then switch to play mode, and then change step the new position of the objects is not saved.

fwild: I think this is intended behaviour, isn't it? Because otherwise it would override the edit mode resting position? no?

I think we can simplify the behaviour - and accordingly also remove the reset option (and reset on step change always)?

robhillman97: Discussed this with @fwild on slack, looks like the fix would be to save the positions when exiting edit mode.

cawke: done

fwild (approved):

PR #1510, opened 2023-07-14
Incorrect pick and place triggers

### LINKED ISSUE #1221: pick and place: jump to for incorrect placement
Right now it is possible to set a jump target for when the user places in the correct location (which can be used to provide positive feedback in the target step). It would be good, if there is the optional possibility to set a jump target for "wrong" placement, so that feedback for incorrect placement can be given (e.g. "think about this hint and try again" -> audio trigger to jump to the first step again).

robhillman97:
Adds pick and place triggers for incorrect placement closes #1221


fwild:
looks good to me?

fwild (approved):

cawke (approved):

PR #1520, opened 2023-07-21
pickandplace: fix saving after exiting edit mode

cawke:


fwild (changes requested):
I don't think this change does what we want it to do - doesn't this mean, it saves in playmode? That should not be the case - because this overwrites the editmode saved position.

cawke:
> I don't think this change does what we want it to do - doesn't this mean, it saves in playmode? That should not be the case - because this overwrites the editmode saved position.

> Rob: Discussed this with @fwild on slack, looks like the fix would be to save the positions when exiting edit mode.

```
        private void OnEditModeChanged(bool editModeState)
        {
            EditModeChanges(editModeState);

            if (!editModeState)    //changed from EditMode to PlayMode
            {
                EventManager.ActivitySaved();
            }
        }
```
position will be saved when you exit edit mode as rob wrote.

fwild:
Ok, got it - so this fix should fix something, and I therefore advocate. As this fix was already in the last preview build, this means, however, that there still is a bug somewhere, which overwrites the stored activity model position when doing the pick&place correct or incorrect placement in playmode. Maybe this has to do with resume state? And we need to reset the resume state?

fwild (approved):

PR #1538, opened 2023-08-07
added watson trigger for any step

### LINKED ISSUE #1436: Watson to trigger any step
Currently it is possible to trigger the next step via Watson. We should improve this so we can trigger any step.

robhillman97:
closes #1436


fwild (changes requested):
revise to parse
%%trigger:step=12%%

fwild (approved):

PR #1640, opened 2023-10-16
Add caption to image augmentation

shidende:
This is adding captions to the image augmentation.


fwild:
One thing I notice is that the data is not stored in the Activity.cs data model - this should be fixed, and maybe, @robhillman97, you could help with that? I think accessibility would deserve a 'caption' attribute to the JSON for images (or, better, we should probably call it 'alt', as in html?).

robhillman97:
Hi Deogratias,

I have pulled the updates from develop and resolved a merge conflict. (my first commit)

I have also changed the caption code so that captions are stored in the data model instead of a text file. The original set up would throw an error at line 195 in FloatingImageViewer.cs whenever an image was created without a caption. We should allow people to not add captions to images if they don't want to so I have removed this. I have also set the caption text object to be disabled by default and only enabled when a caption has been added.  (Second Commit)

There are a couple of other issues:

There are missing scripts/components on the new ImageCaptionEditor_v2 and ImageCaptionPreview_v2 prefabs which prevent the add caption panel from opening/working see below.

ImageEditorObMissingComponents

ImageCaptionEditorMissingScript

The Image editor UI does not stretch to fit the screen anymore see below

Editor does not stretch

We should create a design ticket for this to improve the way captions look (see below). Maybe we could use the Label augmentation for now? @fwild @fominykh 

CaptionExample

Currently there is no implementation for adding captions on the HoloLens UI.

fominykh:
@robhillman97 and @shidende you can check Figma. We have the design for Image captions for the phone UI.

shidende:
Hi @fominykh, I checked the Figma again, it contains the image augmentation design for the authoring side. The issue @robhillman97 has raised is when you have completed adding the image augmentation, and image is hanging in the space (player side?). I have not seen the design of this hanging image together with the caption in the figma.

fominykh:
On smartphone, the captions should always in the same space - in screen space (not together with the image augmentation in the world space). For an image augmentation, the caption will appear in screen space as soon as the image augmentation is visible in the world space i.e. as soon as the user looks at the image.

fominykh:
The only UI we don't have is the Hololens UI.
@shidende if you still need a guide to the caption UI, let me know and we can schefuke a call.

fwild:
@shidende: can you please pull the branch with Rob's fixes, and add the two missing prefab files, who seem currently not to be in the git?

fwild:
@shidende then we can include it in 2.5!

fwild:
Please remove the changes to the .yml files, and check all others changes from your last commit if they were intended!

fwild:
This is still faulty, as the changes were already committed and your latest push has not restored them.
Please do the following (no idea how to do this without commandline):
* open a git bash window
* ```git checkout develop -- .github/FUNDING.yml```
* ```git checkout develop -- .github/workflows/linux_workflow.yml```
* ```git checkout develop -- .github/workflows/.github/workflows/unity_activation_linux.yml```
* ```git checkout develop -- .github/workflows/.github/workflows/unity_activation_windows.yml```
* ```git checkout develop -- .github/workflows/.github/workflows/windows_workflow.yml```
* ```git add .github/*```
* ```git commit -m "restoring workflows and funding"```
* ```git push -f```

Not quite sure if you have to install the commandline tools for that first - gitforwindows has a built in commandline ('git bash'), which should allow you to do exactly the above. I think you can directly access it by right clicking on the MIRAGE-XR folder and selecting 'git BASH'.

PR #1662, opened 2023-11-02
Library tab implimentation

### LINKED ISSUE #1572: Implement model library for 3D model augmentation
The model library is already foreseen in the UI, but not yet implemented.
Two libraries were created: search for BZB23 and search for ARIDLL on sketchfab to see all of them.
This requires: https://github.com/WEKIT-ECS/MIRAGE-XR/issues/1519
As interim solution, we could host the two assetbundles on wekit-ecs.com, and we freed up the space already.

jafarigit:
The model library with two categories (Foods and tools) has been implemented.

Closes #1572


fwild (commented):
Please remove the folder 'codemonster jafari' :) and put the files where they belong?

fwild (commented):
oh, and please remove the copyright header (you can leave a notice that this was authored by you, of course), but the project is governed by the central MIT license:

// -----------------------------------------------------------------------
// Copyright (c) 2023 Codemonster Jafari (Abbas Jafari)
// All rights reserved.
// 
// This software and associated documentation files (the "Software") are 
// provided for use in the project "The silent mystery" and associated works.
// Redistribution, modification, or use in other projects without express 
// written permission from the copyright holder is strictly prohibited.
// -----------------------------------------------------------------------

fwild (commented):
Could we also clean up some spelling mistakes:
Baguete -> Baguette (also for the textures)
Bannana -> Banana
BagueteAO -> BaguetteAO
BagueteNormal -> BaguetteNormal
Yougurt -> Yogurt (also for the texture and asset)
FoodsCatergory -> FoodCategory

cawke (changes requested):
Now the model library window opens by default. local models should open.

image

jafarigit:
> Now the model library window opens by default. local models should open.
> 
> image

Because the model library should be available without login too. If local be the default the login popup opens and doesn't let the user use models before login

fwild:
I was thinking about that, and I am still not sure. Local models, btw, should not require login, as it just offers all the already downloaded models. I think both have their pros and cons. Local models makes it easier to work with content already downloaded, which can happen easily, when you need to delete arrangements and need to re-do them. Library is easier when creating new content from scratch - over time it could become our preferred method to offer content. So: I am happy with either. Maybe Mikhail has a different stance?

fominykh:
I think local models should open by default.

cawke (approved):

PR #1707, opened 2023-12-14
new calibration ui and small fixes

cawke:


fwild:
Did you move all those model library files around? They were not in the right place...
(those hundreds of files listed in this PR under Player/Resources/Icons/Library/)

cawke:
> Did you move all those model library files around? They were not in the right place... (those hundreds of files listed in this PR under Player/Resources/Icons/Library/)

for them I only changed the compression
image

I moved the models for the libraries from the folder with the calibration managers (there should be scripts there, I don’t understand who put them there and why) see the screenshot below.
image

fwild (approved):

fwild:
perfect, just wanted to double check that this did not happen by mistake!

PR #1723, opened 2024-01-04
1689 fixes MoodleManager hit counter error

### LINKED ISSUE #1689: MoodleManager: tries to increase hit counter on server for local activity
Moodle Manager throws an error on the console log when trying to increase the hit counter on the server for an activity that is not yet uploaded, and only locally saved. Expected behaviour: should check if the activity is uploaded before attempting to increase the hit count? Or simply not throw an error.

fwild:
MoodleManager tried to increase the hit counter also for local activities, throwing an error. 

Fixes #1689 by adding a boolean flag to the UpdateViewsOfActivity method to indicate whether the activity existsRemotely.

### DIFF
--- a/Assets/MirageXR/Services/Moodle/MoodleManager.cs
+++ b/Assets/MirageXR/Services/Moodle/MoodleManager.cs
@@ -215,7 +215,6 @@ namespace MirageXR
 
             //Comented out the below debug log due to the size of the message
             //Debug.LogDebug(response);
-
             return ParseArlemListJson(response);
         }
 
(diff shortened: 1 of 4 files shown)


Assets/MirageXR/Services/Moodle/MoodleManager.cs
```diff
@@ -264,7 +271,13 @@ private static List<Session> ParseArlemListJson(string json)
             }
             catch (Exception e)
             {
-                Debug.LogError($"[MoodleManager] ParseArlemListJson error\nmessage: {e}");
+                RootView_v2.Instance.dialog.ShowMiddle(
+                    "Network Error",
+                    "Could not connect to server, showing local activities only.",
+                    "OK", () => AppLog.LogInfo("[MoodleManager] User acknowledged network error"),
+                    "Cancel", () => AppLog.LogInfo("[MoodleManager] User acknowledged network error"),
+                    true);
```
cawke:
```suggestion
                RootView_v2.Instance.dialog.ShowMiddle(
                    "Network Error",
                    "Could not connect to server, showing local activities only.",
                    "OK", () => AppLog.LogInfo("[MoodleManager] User acknowledged network error")
                    true);
```

the dialog now supports a one-button window

cawke (changes requested):

PR #1800, opened 2024-03-07
#1720 sentry

### LINKED ISSUE #1720: Sentry: upgrade to full package
If we are happy with Sentry, we should use the official package (instead of the open source minimal wrapper currently implemented): https://docs.sentry.io/platforms/unity/

fwild:
Adds full Sentry support, resolves #1720


fwild:
I have added now the support for context reporting (user info, session, action id) - and tested it with a second account. this is ready to be reviewed/merged.

Issue #1704, opened 2023-12-07
floor line: 90 degree angle


fwild:
Replace the bezier floor line with a two-segment straight line, so that the line goes from aura to 90 degrees below the task station, and then goes straight up to the task station.

fwild:
And add a toggle switch to the profile so that this feature can be switched off per activity (= activity author's choice, not user choice). I am not sure I really like floor lines, when designing activities that take place on my desk: The floor level is needed to lock the aura in, and anything below floor level becomes unmovable anyways, so also re-setting the floor level to my desk does not really work. When connecting the aura with the diamonds that are floating over my desk (without occlusion handling), it always looks just weird. So: best to have a toggle switch in the activity that allows to switch off the floor lines.

fwild:
replaced by https://github.com/WEKIT-ECS/MIRAGE-XR/issues/1881

Issue #940, opened 2022-06-07
Improve visuals of aura and task station line


fwild:
If the aura had occlusion handling for your feet (or anything above the ground plane), it would look ab it cooler. Adding an alpha channel value with more transparency would make it look better, too. The floor lines could be more like a ribbon over the spatial map, or at least a stronger bezier curve so it runs straight on the floor in front of the user first, before linking to the task station. When the task station is close to the user, it often goes above the diamond, occluding the diamond.

fwild:
superseded by other tickets. closing.

PR #1805, opened 2024-03-08
1751 line paths off and 2nd task station off 

### LINKED ISSUE #1751: Remove next step blue task station diamond

fwild:
Switched visibility of 2nd task station off and switched path to current task station to off. 

The code is still there, in three files:
* TaskStationStateController.cs - line 68ff  (else if isNext)
* PathRoleController.cs - line 92ff (else if isNext)
* WorkPlaceObjectFactory.cs - three times: in CreateThings, CreatePerson, and CreatePlaceObject

These should be removed once we are happy with the removal.
There is probably also a simplification possible with the way we handle these three (things, persons, places) separately.

Resolves #1751 

Related to #1704 and #940

### DIFF
--- a/Assets/MirageXR/Managers/ObjectFactory.cs
+++ b/Assets/MirageXR/Managers/ObjectFactory.cs
@@ -14,6 +14,7 @@ namespace MirageXR
     public class ObjectFactory : MonoBehaviour
     {
         private static ActivityManager activityManager => RootObject.Instance.activityManager;
+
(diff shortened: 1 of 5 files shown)


fwild:
btw: is the file "TaskStation~.prefab" still needed? It was broken, had no longer existing scripts linked - I fixed it, but I don't think it is in use.

cawke (approved):

PR #1809, opened 2024-03-11
Update to Unity 2022.3

### LINKED ISSUE #1784: Update Unity to 2022 LTS
Check the version numbers required for Apple Vision Pro!

BenediktHensen:
This pull request contains the update to Unity 2022.3 and fixes #1784.

After the update process, I had to switch from the MRTK package dependency to an imported .unitypackage version of the MRTK. The reason for this is the following:

- Due to the Unity update, the minimum supported version of ARFoundation changed from 4.x to 5.x, so it was necessary to update to ARFoundation 5.x in the project. This change includes a restructuring where previously distinct packages were merged into the ARFoundation package.
- There is an error in the MRTK when working with ARFoundation for Android: When targeting the Android build platform, the MRTK package is missing an assembly definition reference to "Unity.XR.Core" for its UnityAR provider.
- If the MRTK package is imported via the Package Manager, no changes can be made to it. Hence, I had to resort to the .unitypackage import of the packages. In the code of the MRTK, I then manually fixed the missing assembly reference.


fwild:
We should probably depart from MRTk anyways in the mid-term? A while back I saw that a prerelease candidate of version 3 was available for MRTk, but not sure we would get more trouble with that than with the current version. Could you review the other PRs, and we merge them in before we all switch this Thursday?

BenediktHensen:
Yes, we should decide whether we want to update to MRTK 3 or use a completely different MR SDK - if we even require something apart from ARFoundation and OpenXR. We should check for which features or development tools we are currently using the MRTK 2 and how we could replace these functionalities.

The MRTK 3 is in a new repository https://github.com/MixedRealityToolkit/MixedRealityToolkit-Unity in an own MRTK organization outside of the Microsoft GitHub-organization and seems to be maintained with regular commits.

Updating from MRTK 2 to MRTK 3 would be as significant of a task as switching to a different SDK. For the MRTK, the underlying concepts and architecture have changed, so there is little resemblance to MRTK 2. This means that all the MR functionality that is now realized with the MRTK 2 would need to be re-modeled and refactored to work again with the MRTK 3.

An alternative might be the XR Interaction Toolkit by Unity https://docs.unity3d.com/Packages/com.unity.xr.interaction.toolkit@2.5/manual/index.html which looks like it could replace the MR input/interaction functionality well.

fwild:
I have copied this comment over into an issue where we can further discuss: https://github.com/WEKIT-ECS/MIRAGE-XR/issues/1813

cawke (approved):

Issue #1765, opened 2024-02-18
broken 3D model from sketchfab triggers gltfast error


fwild:
This model:
https://sketchfab.com/3d-models/wet-intersection-american-road-2da9ef808fff4936addefc86c5315cd4
throws the following error:

```
Failed getting triangles. Submesh topology is lines or points.
```

Seems a broken model, but maybe there is a security check we can do to catch it?

fwild:
I have encountered a similar error, when searching for "Rover" on sketchfab and picking the first model (gray)

fwild:
you can also trigger a glTFast error by changing quickly between two steps (one with a 3D model).

fwild:
(RC2.6): Error still thrown: search for "Wet intersection american road" and select the first one that is nicely zoomed in. Observe the error "Failed getting triangles. Submess topology is lines or points." - the model displays correctly, though, but we should not accept broken models to be included - who knows what problems this can cause?

(RC2.6): switching too fast between an empty step and a step with 3D model (or back) still causes a threading exception.

PR #1853, opened 2024-03-14
broken 3D model from sketchfab triggers gltfast error

cawke:
https://github.com/WEKIT-ECS/MIRAGE-XR/issues/1765

image

in case of a broken mesh, when adding MeshCollider (mc), mc will throw an exception. since wrapping it in 'try catch' will not work.

even if you just try to get “triangels” from the mesh, using “try catch” will not work because an error inside the method will be written to the log and the exception will not be thrown.

I added a check and reduced many errors to one, but I can’t completely get rid of them.

### DIFF
--- a/Assets/MirageXR/ContentTypes/3DModel/Model.cs
+++ b/Assets/MirageXR/ContentTypes/3DModel/Model.cs
@@ -256,7 +256,7 @@ namespace MirageXR
 
             var renderers = model.GetComponentsInChildren<Renderer>();
             _colliders = new List<Bounds>();
-
+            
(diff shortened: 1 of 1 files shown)


fwild:
check mesh.vertexCount check before the mesh.triangles call? if 0, don't add mesh? or don't place game object?

cawke:
> check mesh.vertexCount check before the mesh.triangles call? if 0, don't add mesh? or don't place game object?

**mesh.vertexCount** will not be equal to zero and **mesh.triangles** will be

fwild (approved):

PR #1893, opened 2024-03-27
new action editor view was dropped from Content.prefab

fwild:
thus breaking the add augmentation menu, instead the old glyph editor view was linked...?


cawke (commented):
All references to editors in the develop branch are correct, but it is strange that references to TMP_InpufField are missing. Didn't you have this error?

image

fwild:
Just pulled your merge, and when I check now, both TMP input fields show up as correctly set.

But: Shouldn't element 3 be "action editor v2 (action editor view)"?

fwild:
Weirdly enough, for me, if I leave element 3 as "GlyphEditor (Glyph Editor View)" and then start the app, the Content gameObject under Start / View / MobileUI_v2 / Pages / Content / Activity View / Panel / Content" will have element 3 empty, thus throwing an error when I try to bring up the 'add augmentation' menu.

cawke:
> Just pulled your merge, and when I check now, both TMP input fields show up as correctly set.

I updated Text Mesh Pro and everything is fixed.

> But: Shouldn't element 3 be "action editor v2 (action editor view)"?

Yes, you're right, there's a mistake.

cawke (approved):

PR #1894, opened 2024-03-28
open ai manager (open ai for android/ios)

cawke:


fwild (approved):
Although I would recommend we introduce a base class first for the OpenAIManager - so that Andreas can write an alternative implementation linking against his proxy server. But we can do that for 2.7, as that's when we need it.

PR #1922, opened 2024-04-08
open ai assistant

cawke:


fwild (approved):
looks ok - the next step will be to retire the DialogueService and refactor the CharacterController into VirtualInstructorController

PR #1929, opened 2024-04-08
fixed model animations

### LINKED ISSUE #1928: play model animation
Play embedded animation for sketchfab imports

fwild:
resolves #1928 

To test: 
* search for faw56 and import the readplayer me avatar (= export from mixamo, with an animation attached).
* search for "animation" and pick anything, e.g. dragon

### DIFF
--- a/Assets/MirageXR/ContentTypes/3DModel/Model.cs
+++ b/Assets/MirageXR/ContentTypes/3DModel/Model.cs
@@ -113,11 +113,18 @@ namespace MirageXR
         {
             content.option = ZipUtilities.CheckFileForIllegalCharacters(content.option);
             var loadPath = Path.Combine(RootObject.Instance.activityManager.ActivityPath, content.option, GLTF_NAME);
+
+            ImportSettings _importSettings = new()
+            {
+                AnimationMethod = AnimationMethod.Legacy
+            };
+
             _gltf = new GltfImport();
-            var success = await _gltf.Load(new Uri(loadPath));
+            var success = await _gltf.Load(new Uri(loadPath), _importSettings);
+            var instantiator = new GameObjectInstantiator(_gltf, transform);
             if (success)
             {
-                success = await _gltf.InstantiateMainSceneAsync(transform);
+                success = await _gltf.InstantiateMainSceneAsync(instantiator); // was: transform
             }
 
             if (!success)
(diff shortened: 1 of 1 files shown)


fwild:
@cawke please check whether lines 167-181 are really not needed - they are the culprit that caused the animations not to play (possibly amongst other things: animations were attached to a parent-parent object of where the gltfast importer attaches them already; possibly the import settings needed animation import to be set to legacy; possibly some of the other animation flags were not explicitly set and needed).

Assets/MirageXR/ContentTypes/3DModel/Model.cs
```diff
@@ -126,65 +133,92 @@ private async Task<bool> LoadGltf(ToggleObject content)
...
             var modelTransform = transform;
             var startPos = modelTransform.position + modelTransform.forward * -0.5f + modelTransform.up * -0.1f;
-
+            
             model.transform.SetParent(modelTransform);
 
+            /*
+             * ALEX: please check this - I don't think it is needed and it prevented the animation 
+             * from being played
+             *
```
cawke: Yes, I don't think it's necessary. but we need to understand that models can have different sizes and it would be nice to have a tool that will fit the loaded model into an adequate size (also so that we can specify the original size of the model)

Assets/MirageXR/ContentTypes/3DModel/Model.cs
```diff
@@ -126,65 +133,92 @@ private async Task<bool> LoadGltf(ToggleObject content)
...
+            if (legacyAnimation != null)
+            {
+                Debug.Log("playing animation (with data from gltfast scene instantiator).");
+
+                legacyAnimation.clip = _gltf.GetAnimationClips()[0];
+                legacyAnimation.clip.wrapMode = UnityEngine.WrapMode.Loop;
+                legacyAnimation.clip.legacy = true;
+                legacyAnimation.clip.EnsureQuaternionContinuity();
+                legacyAnimation.playAutomatically = true;
+                legacyAnimation.Play(_gltf.GetAnimationClips()[0].name);
```
cawke: add a null check

cawke (changes requested):

Assets/MirageXR/ContentTypes/3DModel/Model.cs
fwild: I am not quite sure what broke it - there was a warning floating by that said something about "size set to infinite" or some such, so I guess it was simply too small for the animation component to accept.

Assets/MirageXR/ContentTypes/3DModel/Model.cs
fwild: added.

Assets/MirageXR/ContentTypes/3DModel/Model.cs
fwild: I checked again, and it turns out the mode renaming was the real culprit, so I restored all the resize functions and only commented this out:
```
// model.name = _obj.option;
```
Is this model.name relevant anywhere? I think not, right?

cawke: I guess.

Assets/MirageXR/ContentTypes/3DModel/Model.cs
```diff
@@ -126,22 +133,40 @@ private async Task<bool> LoadGltf(ToggleObject content)
...
+                if (clips.Length > 0)
+                {
+                    legacyAnimation.clip = clips[0];
+                    legacyAnimation.clip.wrapMode = UnityEngine.WrapMode.Loop;
+                    legacyAnimation.clip.legacy = true;
+                    legacyAnimation.clip.EnsureQuaternionContinuity();
+                    legacyAnimation.playAutomatically = true;
+                    legacyAnimation.Play(clips[0].name);
+                }
+            }
```
cawke:
```suggestion
                var clips = _gltf.GetAnimationClips();
                if (clips != null && clips.Length > 0 && clips[0] != null)
                {
                    legacyAnimation.clip = clips[0];
                    legacyAnimation.clip.wrapMode = UnityEngine.WrapMode.Loop;
                    legacyAnimation.clip.legacy = true;
                    legacyAnimation.clip.EnsureQuaternionContinuity();
                    legacyAnimation.playAutomatically = true;
                    legacyAnimation.Play(clips[0].name);
                }
            }
```

cawke (changes requested):

Assets/MirageXR/ContentTypes/3DModel/Model.cs
fwild: I checked carefully - deleting, adding, keep alive, upload/download, all works. Merging.

PR #1934, opened 2024-04-09
small bugfix

Alvareon:


fwild:
merged, but still has bug: search field for sketchfab no longer selectable, local models not scrollable.

PR #1937, opened 2024-04-10
Silyasov/asset bugfix

fwild:
I re-added the sketchfab token prefabs to the mobile editor panel prefab of the model augmentation. At least for me, these were dropped (and usually wouldn't be?).


cawke:
We have these scriptable objects in gitignor and that is why references to them are broken because we can't store keys in git. we need to rewrite the way we get keys to them and think about some key manager that will take care of all this.

PR #1944, opened 2024-04-15
camera render fix

cawke:
image

### DIFF
--- a/Assets/MirageXR/MirageXR.asmdef
+++ b/Assets/MirageXR/MirageXR.asmdef
@@ -27,7 +27,8 @@
         "GUID:a42927d1d4a3b4cda9b076a7adecb9cc",
         "GUID:dc960734dc080426fa6612f1c5fe95f3",
         "GUID:3248779d86bd31747b5d2214f30b01ac",
-        "GUID:7958db66189566541a6363568aee1575"
+        "GUID:7958db66189566541a6363568aee1575",
+        "GUID:15fc0a57446b3144c949da3e2b9737a9"
     ],
     "includePlatforms": [],
     "excludePlatforms": [],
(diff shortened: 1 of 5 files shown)


BenediktHensen (commented):
Unfortunately, this only works in portrait mode but not in landscape orientation. This is also were I am currently stuck.

When the device is in landscape orientation, the overlay gets stretched across the entire screen instead of positioning it on the right side. It also seems to mess up the input mapping as I can no longer click any of the buttons. Instead, their interaction area is somewhere else on the screen but not, where the visible button is.
grafik

Apparently, there is a limitation in URP that overlays cannot be moved and scaled using the viewport rect, so our previous method of positioning the UI does not work anymore. https://forum.unity.com/threads/why-cant-overlay-cameras-use-viewport-rect.924932/

The suggested solution in the forum is to use a render texture but that seems suboptimal.
Maybe camera projection matrices could help by directly manipulating the overlay camera image into position but that could also mess up the mapping of input coordinates when clicking on the buttons.
Another solution might be to adjust the entire UI so that it always covers the entire screen. For landscape resolutions, we could then add a blank/transparent UI element on the left which pushes the actual UI content over to the right but that is quite a substantial change and could also interfere with inputs that should reach the objects in the scene.

BenediktHensen:
So it seems like this line here does not have an effect anymore in the URP pipeline.

https://github.com/WEKIT-ECS/MIRAGE-XR/pull/1944/files#diff-72fd6680344304866a797542739b89d7fa2ec3974810953546c5a4eb718922e1R60

fwild:
Yes, I get that too: 

image

cawke:
Curious. I didn't see that. I think the PR can be closed.

Issue #1904, opened 2024-04-02
remove mcs.rsp 


fwild:
It seems the old mono compiler custom response file is still in the project, but obsolete -- see https://forum.unity.com/threads/help-using-obsolete-custom-response-file-mcs-rsp-please-use-csc-rsp-instead.1072442/

fwild:
The only thing it contains seems to be legacy:

```
-r:System.IO.Compression.dll
-r:System.IO.Compression.FileSystem.dll
```

And here is a way to migrate this to csc.rsp: https://forum.unity.com/threads/c-compression-zip-missing.577492/page-2

fwild:
tested it, and it does not seem to need this statement, nor the file.

fwild:
might not be related, but in my test branch I get now the warning that 
```
Script 'UI_KeyboardInputField' on GameObject TitleInputField has possibly missing Required Components!
```
and the script has some directives in there that only apply if Unity version bigger than 2019...

fwild:
after long investigation, I am unsure whether this file (and r-include) is needed

PR #1950, opened 2024-04-19
URP updates for Android build

fwild:
* added missing feature setting for AR camera in URP profile
* added network capability for android manifest
* set FXAA as anti-aliasing as TAA was not valid for Android
* added missing meta file msc.rsp.meta (for compiler flags already in the project)

This resolves the issue that the camera shows up as black on Android builds
and that it can reach the Moodle server again.


fwild:
Resolves #1952 and #1953

Assets/msc.rsp.meta
BenediktHensen, line 1: We wanted to get rid of the msc.rsp file (see #1904) and so, the file, along with its meta file, were removed in an earlier pull request. However, some time after that, we ended up in a strange in-between stage where the msc.rsp file re-appeared but its meta file is still deleted. So, now Unity keeps adding a new meta file for the file on every new branch.

This doesn't block this pull request but we should ensure after merging this that we get back into a good state where the msc.rsp file is completely deleted.

Assets/MirageXR/Rendering/UniversalRenderPipelineGlobalSettings.asset
BenediktHensen, line 1: Is Unity stripping away too many shader variants so that we need to disable this optimization?

In the future, we should look into methods to define exactly which shader variants are required so that we can re-enable stripping the unused variants.

BenediktHensen (approved):

Assets/MirageXR/Rendering/UniversalRenderPipelineGlobalSettings.asset
fwild: good point - I had added this, as we lost the shadows AND the 3d model sketchfab imports. Not actually sure this is needed!

PR #1961, opened 2024-04-23
Fwild/urp materials

### LINKED ISSUE #1951: gltf shader materials missing
It seems the shader materials for gltf URP are not included in the build, thus throwing an exception whenever a 3D model is imported from sketchfab.

fwild:
Resolves https://github.com/WEKIT-ECS/MIRAGE-XR/issues/1951

(But let me double check this first!)


fwild:
(this PR has messed up commits - will clean later!)

fwild:
Cleaned it up - ready to go.

cawke:
I think we should leave only the basic shaders and not collect all possible shader variants. even on a 12-core processor it takes a lot of time (I understand that this is only the first time) and I’m afraid to imagine what size of the build this assembly will have.

image

cawke:
image
The build is over 400 megabytes in size. which is more than double the normal build.

fwild:
This is now a small build, and it seems all problems are gone.

PR #1970, opened 2024-04-24
1806 task station diamond make configurable

### LINKED ISSUE #1806: Task station diamond: make configurable
Often users place the diamond too far away or too close - why not make the diamond configurable with a toggle, also allowing to set its size, or maybe offer a couple of alternative visualisations (portal, flat floor marker launch symbol, flat book marker launch symbol)?

Related to: https://github.com/WEKIT-ECS/MIRAGE-XR/issues/1539

AKissmehl:
Diamond is now configurable. Please take a look at the changes in Open XR Package Setting.


fwild (changes requested):
I tested it, but for me the toggle has no effect? I don't see any error message either?
Also: please roll back the files from the commit that are not relevant - you changed a lot of packages, other prefabs (like ghosteditor), a PR should only contain what has been changed!

PR #1998, opened 2024-05-28
1806 diamond visibility

### LINKED ISSUE #1806: Task station diamond: make configurable

cawke:
https://github.com/WEKIT-ECS/MIRAGE-XR/issues/1806

### DIFF
--- a/Assets/MirageXR/DataModel/Activity.cs
+++ b/Assets/MirageXR/DataModel/Activity.cs
@@ -23,6 +23,12 @@ namespace MirageXR
     }
 
     [Serializable]
+    public class ActionExtension
+    {
+        public bool isDiamondVisible;
+    }
+
+    [Serializable]
     public class Action // is a task station
     {
         // NOT IN ARLEM SPEC. For storing current action state.
--- a/Assets/MirageXR/DataModel/Activity.cs
+++ b/Assets/MirageXR/DataModel/Activity.cs
@@ -23,17 +23,12 @@ namespace MirageXR
     }
 
     [Serializable]
-    public class ActionExtension
-    {
-        public bool isDiamondVisible;
-    }
-
-    [Serializable]
     public class Action // is a task station
     {
         // NOT IN ARLEM SPEC. For storing current action state.
         public bool isActive;
         public bool isCompleted;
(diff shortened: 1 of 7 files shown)


Assets/MirageXR/DataModel/Activity.cs
```diff
@@ -22,6 +22,12 @@ public class Activity
         public List<Action> actions = new List<Action>();
     }
 
+    [Serializable]
+    public class ActionExtension
+    {
+        public bool isDiamondVisible;
```
fwild: Question: why add this as own class ActionExtension - instead of simply adding it as a bool (like "isActive" or "isCompleted") inside of the class "action" below?

Assets/MirageXR/Scripts/CombinedEditor/TaskStationEditor.cs
```diff
@@ -12,52 +14,37 @@ public class TaskStationEditor : MonoBehaviour
 
     private ObjectManipulator _objectManipulator;
     private TaskStationStateController _taskStationStateController;
-    private PlaceBehaviour _placeBehaviour;
+    private MeshRenderer _meshRenderer;
     private Detectable _detectable;
-    private bool _isInitialized;
```
fwild, line 17: I see that "_isInitialized" is removed - I think this was used to flag of the activity was calibrated? So this is no longer necessary?

fwild: Resolved after discussion: this is just used internally to prevent calling initialisation multiple times (for whatever reason)

Assets/MirageXR/DataModel/Activity.cs
fwild: @cawke?

cawke: updated

PR #1999, opened 2024-05-29
Tutorial refactor

BorisJov:


Assets/MirageXR/Tutorial/HelpSteps/HelpSelectionModel.cs
```diff
@@ -0,0 +1,51 @@
...
+    /// <summary>
+    /// Model for all items in the help selection dialogue.
+    /// Should be serialisable.
+    /// </summary>
+    public class HelpSelectionModel
+    {
+        /// <summary>
+        /// Text that should be shown in the help popup, when the user is selecting help.
+        /// </summary>
+        [JsonProperty("selection_text")] public string SelectionText { get; set; }
```
cawke:
```suggestion
        [JsonProperty] public string SelectionText { get; set; }
```

for others too

Assets/MirageXR/Player/Scripts/Mobile/Popup/HelpSelectionPopup.cs
```diff
@@ -41,5 +143,28 @@ protected override bool TryToGetArguments(params object[] args)
...
+        /// </summary>
+        /// <param name="steps">The steps in the mini-tutorial.</param>
+        private void ShowShortMessageSequence(List<TutorialModelUI> steps)
+        {
+            this.Close();
+            var queue = new Queue<TutorialModelUI>(steps);
+            TutorialManager.Instance.MobileTutorial.Show(queue);
+        }
+
+        private void startMobileEditingTutorial()
```
cawke:
```suggestion
        private void StartMobileEditingTutorial()
```

Assets/MirageXR/Player/Scripts/Mobile/Popup/HelpSelectionPopup.cs
```diff
@@ -41,5 +143,28 @@ protected override bool TryToGetArguments(params object[] args)
...
+        {
+            this.Close();
+            TutorialManager.Instance.StartNewMobileEditingTutorial();
+        }
+
+        private void startMobileViewingTutorial()
```
cawke:
```suggestion
        private void StartMobileViewingTutorial()
```

cawke (changes requested):
generally good, correct minor errors and check the positions of all windows so that the situation does not occur as in the screenshot

image

Assets/MirageXR/Tutorial/HelpSteps/HelpSelectionModel.cs
BorisJov: Why remove these? How will it parse then?

BorisJov: I've just tried removing the names and it doesn't work without them.

PR #2003, opened 2024-06-04
Fixing the caption field on top of other UI

shidende:
This merge wants to fix the appearance of the caption field on the top of other UI when the audio prefab is active, but the caption is not generated. Now, the caption will be displayed only when it is generated.

### DIFF
--- a/Assets/MirageXR/ContentTypes/Audio/AudioPlayer.cs
+++ b/Assets/MirageXR/ContentTypes/Audio/AudioPlayer.cs
@@ -117,19 +117,26 @@ namespace MirageXR
             }
 
             var caption = obj.caption;
-            if (caption != string.Empty)
-            {
+            
                 StartCaptionDisplay(caption);
-            }
 
             // If all went well, return true.
             return true;
         }
 
+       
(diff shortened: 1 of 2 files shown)


fwild:
I have tested this branch, but I think it is not working correctly? when I start playmode, create a new activity (with audio and with caption), and then look in the hierarchy for "caption", I can find a lot of gameobjects - some attached to the Aura for hololens, seems like they switch on and off the captions? But on mobile (build target: Android), the captions are not displayed. In fact, the captionOn gameObject in the content is switched off, and the profile setting button does not keep its value? Moreover, the text field inside of the captionOn is empty? Not quite sure how this is supposed to work, but it seems not to work for me?

fwild:
Deo, we had to roll back the changes as it made the code break, and as it was not functioning. We have released now version 2.7, and you can restart your pull request - but please test that it is working before hands. Could you also please add wiki documentation to explain how it is implemented? I propose to close this PR, and start a new one...

fwild:
After communication, I am closing this PR - please start a new one when resubmitting!

Issue #1995, opened 2024-05-23
Migrate CharacterModel logic to VirtualInstructor


fwild:
The editor panels are implemented, but the player code not yet migrated (and tied to the changes in data model).

- [x] check UI (and add design proposal for push-to-talk button)
- [x] animations: hard coded idle animation, maybe add gesticulating when talking
- [ ] captions missing (and need to be improved over old character model, also add system setting captions on/off)
- [ ] follow the user (with improvement that it positions slightly in front of the user to the side)
- [ ] other missing logic

PR #2009, opened 2024-06-07
1995 migrate charactermodel logic to virtualinstructor

AKissmehl:
A first version of the Virtual Instructor. 

Addresses: https://github.com/WEKIT-ECS/MIRAGE-XR/issues/1995


fwild (approved):
Looks good from my side

Issue #2012, opened 2024-06-13
TMR: Serialising existing tutorials and refactoring system


BorisJov:
As was done for HelpSelection in #1991 we need to store tutorials in JSON and refator the tutorial system accordingly.
This will be a core change for the TutorialManager, also changing how steps are processed and modelled.

PR #2014, opened 2024-06-18
TMR: Serilasing tutorials

BorisJov:
Mobile tutorials are now using the new JSON-serialised system.
The rest of the existing tutorials (Hololens) are waiting on the new UI to be refactored.
Closes issue #2012 for now.


cawke (changes requested):
As far as I understand, you did not delete old unused classes. We also did not remove unused events in the event manager

image

BorisJov:
Yes, exactly. I intended to remove all of them once the tutorial is fully refactored. This is dependent on the new Hololens UI design, for which we will do tutorial later in the year. Especially for the events I would prefer to do it this way, so as not do the same job twice.

cawke (approved):
ok

PR #2040, opened 2024-07-17
basic build pipeline

fwild:
A first basic build pipeline - just with a regular android build, not quest3 yet


cawke:
to master?

Assets/MirageXR/Automation/Editor/LocalBuildPipeline.cs
```diff
@@ -0,0 +1,116 @@
...
+            {
+                Debug.Log("Build succeeded: " + summary.totalSize + " bytes");
+            }
+
+            if (summary.result == BuildResult.Failed)
+            {
+                Debug.Log("Build failed");
+            }
+
+        }
```
cawke, line 113: please fix the formatting

cawke (changes requested):

Assets/MirageXR/Automation/Editor/LocalBuildPipeline.cs
```diff
@@ -0,0 +1,116 @@
...
+	/// <summary>
+	/// This class adds new menu items to the editor for compiling MirageXR for specific platforms.
+	/// </summary>
+	public class LocalBuildPipeline
+    {
+
+        /// <summary>
+        /// List of scenes to include in the build
+        /// </summary>
+        public static string[] Scenes = { "Assets/MirageXR/Scenes/Start.unity", "Assets/MirageXR/Scenes/Player.unity", "Assets/MirageXR/Scenes/ActivitySelection.unity" };
```
BenediktHensen, line 23: We could use `EditorBuildSettings.scenes` https://docs.unity3d.com/ScriptReference/EditorBuildSettings-scenes.html instead of hardcoding the scene paths, as this is probably prone to breaking if we ever reorganize the folder structure. Then, the build pipeline would automatically take the scenes that are configured in the Build Window. If someones adds a new scene to the build in the future as they will probably overlook that the list in this script would need to be updated, too.

BenediktHensen (commented):
The build works and it creates an apk file. I think everything should be fine with it, functionality-wise. One suggestion is to automatically fetch the included build scenes so that there still is a single source of truth to avoid confusion later on if we change the scenes in the build.

Initially, there were some Unity bugs and crashes with this branch, as Unity somehow did not initialize the buildtargetSettings.AssignedSettings reliably and it was always null, leading to a NullReferenceException when calling the XRPackageMetadataStore.AssignLoader. Sometimes, it would then initialize but then lead to crashes and corrupted XR Settings files during the build. However, after deleting the Library file and letting Unity re-generate the project, the pipeline now works reliably, so it was likely some old cache file that I had in the project.

Assets/MirageXR/Automation/Editor/LocalBuildPipeline.cs
fwild: done

PR #2045, opened 2024-07-25
New data model

cawke:


Assets/MirageXR/DataModel/DataModel.cs
```diff
@@ -0,0 +1,319 @@
...
+    {
+        [JsonProperty] public Guid Id { get; set; }
+        [JsonProperty] public DateTime CreationDate { get; set; }
+        [JsonProperty] public string Version { get; set; }
+    }
+
+    [Serializable]
+    public class Activity : DataModel
+    {
+        [JsonProperty] public User Owner { get; set; }
```
fwild: I think this should not be inside the data model, also as ownership is m:n now with collaborative authoring. If we wanted to we could rename to 'creator' or 'created_by', but I think this needs to be handled by the repository outside of the activity data model.

Assets/MirageXR/DataModel/DataModel.cs
```diff
@@ -0,0 +1,319 @@
...
+    }
+
+    [Serializable]
+    public class Media
+    {
+        [JsonProperty] public List<File> File { get; set; } //one or many?
+    }
+
+    [Serializable]
+    public class RequiredTools
```
fwild: I think this should be "RequiredToolsPartsMaterials" or maybe short "RequiredTPM"? and add "Tools/Parts/Materials as comment to explain?

Assets/MirageXR/DataModel/DataModel.cs
```diff
@@ -0,0 +1,319 @@
...
+    {
+        [JsonProperty] public List<string> Tools { get; set; }
```
fwild: again "ToolPartMaterial" or maybe short "TPM"?

Assets/MirageXR/DataModel/DataModel.cs
```diff
@@ -0,0 +1,319 @@
...
+    public class ActivityStep : DataModel
+    {
+        [JsonProperty] public uint Order { get; set; } //or it can be based on Activity.Steps?
+        [JsonProperty] public string Name { get; set; }
+        [JsonProperty] public string Description { get; set; }
+        [JsonProperty] public List<Guid> Contents { get; set; } //Content.Id
+        [JsonProperty] public Location Location { get; set; }
+        [JsonProperty] public List<Trigger>? Triggers { get; set; } // cannot be null?
+        [JsonProperty] public Comments Comments { get; set; }
+        [JsonProperty] public RequiredTools RequiredTools { get; set; }
```
fwild: Should be `List<RequiredTPM> RequiredTPMs`?

Assets/MirageXR/DataModel/DataModel.cs
fwild: And why is this one a class, whereas the Triggers are directly included in the ActivityStep as list?

Assets/MirageXR/DataModel/DataModel.cs
```diff
@@ -0,0 +1,319 @@
...
+    {
+        [JsonProperty] public User User { get; set; }
+        [JsonProperty] public string Text { get; set; }
+        [JsonProperty] public Media? Attachment { get; set; }
+    }
+
+    [Serializable]
+    public class Media
+    {
+        [JsonProperty] public List<File> File { get; set; } //one or many?
```
fwild: should be `List<File> Files` for consistancy

Assets/MirageXR/DataModel/DataModel.cs
```diff
@@ -0,0 +1,319 @@
...
+    {
+        [JsonProperty] public uint Order { get; set; } //or it can be based on Activity.Steps?
+        [JsonProperty] public string Name { get; set; }
+        [JsonProperty] public string Description { get; set; }
+        [JsonProperty] public List<Guid> Contents { get; set; } //Content.Id
+        [JsonProperty] public Location Location { get; set; }
+        [JsonProperty] public List<Trigger>? Triggers { get; set; } // cannot be null?
+        [JsonProperty] public Comments Comments { get; set; }
+        [JsonProperty] public RequiredTools RequiredTools { get; set; }
+        [JsonProperty] public Media Media { get; set; }
```
fwild: Why not `list<MediaAsset> MediaAssets`?

Assets/MirageXR/DataModel/DataModel.cs
```diff
@@ -0,0 +1,319 @@
...
+    {
+        [JsonProperty] public List<Comment> Note { get; set; }
+    }
+
+    [Serializable]
+    public class Comment : DataModel
+    {
+        [JsonProperty] public User User { get; set; }
+        [JsonProperty] public string Text { get; set; }
+        [JsonProperty] public Media? Attachment { get; set; }
```
fwild: Again not sure why this could not be directly a `list<MediaAsset> Attachment` - for consistancy

Assets/MirageXR/DataModel/DataModel.cs
```diff
@@ -0,0 +1,319 @@
...
+    [Serializable]
+    public class RequiredTools
+    {
+        [JsonProperty] public List<string> Tools { get; set; }
+    }
+
+    [Serializable]
+    public class User : DataModel
+    {
+        [JsonProperty] public string Name { get; set; }
```
fwild, line 79: `public string UUID` instead? To be more precise? We cannot maintain a full list of users locally - if we need to find e.g. a clear name for a user, we will have to query the identity server for properties we are allowed to access. As far as I can see, the unique user ID (UUID) is the way to go with keycloak for the identification of a user.

Assets/MirageXR/DataModel/DataModel.cs
```diff
@@ -0,0 +1,319 @@
...
+                Version = Version,
+                Type = Type,
+                ContentData = ContentData.ToObject<T>(),
+                Location = Location
+            };
+        }
+    }
+
+    [Serializable]
+    public class File : DataModel
```
fwild, line 107: I would propose NOT to have our own File class, but use System.IO.File instead. Unless there is a better existing File class from a different library. Rationale is: if we use the standard File interface, we can directly open a StreamReader etc.

Assets/MirageXR/DataModel/DataModel.cs
```diff
@@ -0,0 +1,319 @@
...
+    [Serializable]
+    public class Content<T> : DataModel where T : ContentData
+    {
+        [JsonProperty] public ContentType Type { get; set; }
+        [JsonProperty] public string Name { get; set; }
+        [JsonProperty] public Location Location { get; set; }
+        [JsonProperty] public T ContentData { get; set; }
+    }
+
+    public class Trigger    //todo
```
fwild, line 142: Missing `[Serializable]`

Assets/MirageXR/DataModel/DataModel.cs
```diff
@@ -0,0 +1,319 @@
...
+    {
+        [JsonProperty] public TriggerType Type { get; set; }
+        [JsonProperty] public Guid StepId { get; set; } // step.id
+        [JsonProperty] public string Event { get; set; } // string needs to be eval’d
+        [JsonProperty] public string PromptInjectionText { get; set; } // What text to drop to the AI conversation when the trigger is triggered (e.g. “I pick the Banana”, “I have done that”)
+    }
+
+    public enum TriggerType     //todo
```
fwild, line 160: missing `[Serializable]`

Assets/MirageXR/DataModel/DataModel.cs
```diff
@@ -0,0 +1,319 @@
...
+    {
+        Unknown = -1,
+        Next = 1,
```
fwild: let's keep `Click' instead of `Next` - I think this was better

Assets/MirageXR/DataModel/DataModel.cs
```diff
@@ -0,0 +1,319 @@
...
+        Sensor = 2,
+        Detect = 3,
+        Interaction = 4 // maybe? From improved pick&place, that uses events to advance?
```
fwild: Rename to `Event`?

Assets/MirageXR/DataModel/DataModel.cs
```diff
@@ -0,0 +1,319 @@
...
+        [JsonProperty] public Location Location { get; set; }
+        [JsonProperty] public T ContentData { get; set; }
+    }
+
+    public class Trigger    //todo
+    {
+        [JsonProperty] public TriggerType Type { get; set; }
+        [JsonProperty] public Guid StepId { get; set; } // step.id
+        [JsonProperty] public string Event { get; set; } // string needs to be eval’d
+        [JsonProperty] public string PromptInjectionText { get; set; } // What text to drop to the AI conversation when the trigger is triggered (e.g. “I pick the Banana”, “I have done that”)
```
fwild, line 151: for the Sensor trigger to work, we need to still have:
```
 [JsonProperty] public string SensorChannel {get;set}
 [JsonProperty] public string SensorVariable {get;set}
 [JsonProperty] public string SensorVariableType {get;set}
 [JsonProperty] public string SensorExpectedValue {get;set}
 [JsonProperty] public string SensorEvalOperator {get;set} // "exceed", "below", "equal", "between" - or ">", "<", "=", "[x;y]"
```

Assets/MirageXR/DataModel/DataModel.cs
```diff
@@ -0,0 +1,319 @@
...
+    {
+        Unknown = -1,
+        Next = 1,
+        Sensor = 2,
+        Detect = 3,
+        Interaction = 4 // maybe? From improved pick&place, that uses events to advance?
+        //todo
+    }
+
+    public class ContentData
```
fwild, line 172: missing `[Serializable]'

Assets/MirageXR/DataModel/DataModel.cs
```diff
@@ -0,0 +1,319 @@
...
+    {
+        [JsonProperty] public List<Trigger> Triggers { get; set; }
+        [JsonProperty] public List<TriggerType> AvailableTriggers { get; set; }
+    }
+
+    public class Location
```
fwild, line 179: missing `[Serializable]'

Assets/MirageXR/DataModel/DataModel.cs
```diff
@@ -0,0 +1,319 @@
...
+        Sensor = 2,
+        Detect = 3,
+        Interaction = 4 // maybe? From improved pick&place, that uses events to advance?
+        //todo
+    }
+
+    public class ContentData
+    {
+        [JsonProperty] public List<Trigger> Triggers { get; set; }
+        [JsonProperty] public List<TriggerType> AvailableTriggers { get; set; }
```
fwild, line 175: do we need this? what does it do?

Assets/MirageXR/DataModel/DataModel.cs
```diff
@@ -0,0 +1,319 @@
...
+
+    public class Location
+    {
+        [JsonProperty] public Vector3 Position { get; set; }
+        [JsonProperty] public Vector3 Rotation { get; set; }
+        [JsonProperty] public Vector3 Scale { get; set; }
+        [JsonProperty] public TargetMarker? TargetMarker { get; set; } //can be null
+    }
+
+    public class TargetMarker
```
fwild, line 188: missing `[Serializable]'

Assets/MirageXR/DataModel/DataModel.cs
```diff
@@ -0,0 +1,319 @@
...
+        [JsonProperty] public File Audio { get; set; }
+        [JsonProperty] public List<GhostDataFrame> Frames { get; set; }
+        [JsonProperty] public Gender Gender { get; set; }
+        [JsonProperty] public bool IsLooped { get; set; }
+    }
+
+    [Serializable]
+    public class GlyphContentData : ContentData
+    {
+        [JsonProperty] public string FileName { get; set; }
```
fwild, line 269: Shouldn't this be an addressable?

Assets/MirageXR/DataModel/DataModel.cs
```diff
@@ -0,0 +1,319 @@
...
+        [JsonProperty] public string Text { get; set; }
+        [JsonProperty] public File Image { get; set; }
+    }
+
+    [Serializable]
+    public class ModelContentData : ContentData //or 3dModelContentData 
+    {
+        [JsonProperty] public bool isLibraryModel { get; set; }
+        [JsonProperty] public File? Model { get; set; }
+        [JsonProperty] public LibraryModel? LibraryModel { get; set; }
```
fwild, line 284: We could treat the library models the same as the other downloads - and clone/copy plus upload them as separate files into the repository -- as we cannot rely on the library models to not change over time?

fwild: but then we still would have to keep this data, as we need to be able to edit it. got it.

Assets/MirageXR/DataModel/DataModel.cs
```diff
@@ -0,0 +1,319 @@
...
+    [Serializable]
+    public class InteractionContentData : ContentData
+    {
+        [JsonProperty] public string Text { get; set; }
+    }
+
+    [Serializable]
+    public class VfxContentData : ContentData
+    {
+        [JsonProperty] public string FileName { get; set; }
```
fwild, line 313: shouldn't this be `public File file`? To keep consistant with the other assets?

Assets/MirageXR/DataModel/DataModel.cs
```diff
@@ -0,0 +1,319 @@
...
+    [Serializable]
+    public class Comment : DataModel
+    {
+        [JsonProperty] public User User { get; set; }
+        [JsonProperty] public string Text { get; set; }
+        [JsonProperty] public Media? Attachment { get; set; }
+    }
+
+    [Serializable]
+    public class Media
```
fwild: rename to `MediaAsset`, as media is a plural noun (and 'medium' does not sound good)

Assets/MirageXR/DataModel/DataModel.cs
```diff
@@ -0,0 +1,319 @@
...
+    [Serializable]
+    public class Hierarchy
+    {
+        [JsonProperty] public List<HierarchyItem> Item { get; set; }
+    }
+
+    [Serializable]
+    public class HierarchyItem
+    {
+        [JsonProperty] public List<Guid>? StepIds { get; set; }
```
fwild, line 41: Shouldn't this be `ContentIDs`?

Assets/MirageXR/DataModel/DataModel.cs
```diff
@@ -0,0 +1,319 @@
...
+        [JsonProperty] public string Name { get; set; }
+        [JsonProperty] public string Description { get; set; }
+        [JsonProperty] public string Language { get; set; } //is it needed?
+        [JsonProperty] public File Thumbnail { get; set; }
+        [JsonProperty] public Hierarchy Hierarchy { get; set; }
+        [JsonProperty] public List<Content> Content { get; set; } //todo
+    }
+
+    [Serializable]
+    public class Hierarchy
```
fwild: I think `StepHierarchy` would be more intuitive?

Assets/MirageXR/DataModel/DataModel.cs
```diff
@@ -0,0 +1,319 @@
...
+    [Serializable]
+    public class Activity : DataModel
+    {
+        [JsonProperty] public User Owner { get; set; }
+        [JsonProperty] public string Name { get; set; }
+        [JsonProperty] public string Description { get; set; }
+        [JsonProperty] public string Language { get; set; } //is it needed?
+        [JsonProperty] public File Thumbnail { get; set; }
+        [JsonProperty] public Hierarchy Hierarchy { get; set; }
+        [JsonProperty] public List<Content> Content { get; set; } //todo
```
fwild: I think we are missing `public List<ActivityStep> Steps` here? Unless the `ActivityStep` is forgotten in `HierarchyItem`? Though I like it that HierarchyItem has only a reference to the step Guid, but not the data itself, and we would keep the step data here (again: with the ActivityStep's Contents being only Guids and the real content being stored here in Content)

fwild (changes requested):
I added quite a few comments - trying to make the way we use certain data types more consistent. I think I found a mistake: ActivityStep is currently not referenced in Activity, and it should be (seems better than referencing ActivityStep in HierarchyItem).

Assets/MirageXR/DataModel/DataModel.cs
cawke: ++

Assets/MirageXR/DataModel/DataModel.cs
cawke: ++

Assets/MirageXR/DataModel/DataModel.cs
cawke: ++

Assets/MirageXR/DataModel/DataModel.cs
cawke: ++

cawke: no, here are the links to the step

Assets/MirageXR/DataModel/DataModel.cs
cawke: ++

It is better not to use abbreviations in class names

Assets/MirageXR/DataModel/DataModel.cs
cawke: ++

Assets/MirageXR/DataModel/DataModel.cs
cawke: I wasn't sure what should be stored in ToolPartMaterial and that's why I separated it into a class.

Assets/MirageXR/DataModel/DataModel.cs
cawke: ++

Assets/MirageXR/DataModel/DataModel.cs
cawke: ++

Assets/MirageXR/DataModel/DataModel.cs
cawke: List<<File>>? Attachment

Assets/MirageXR/DataModel/DataModel.cs
cawke: ++

Assets/MirageXR/DataModel/DataModel.cs
cawke: I thought to use guid Id from DataModel

Assets/MirageXR/DataModel/DataModel.cs
cawke: I don't think we can store an instance of System.File in the data model

cawke: all tasks for downloading and checking the file will be on another manager.

Assets/MirageXR/DataModel/DataModel.cs
cawke: ++

Assets/MirageXR/DataModel/DataModel.cs
cawke: ++

Assets/MirageXR/DataModel/DataModel.cs
cawke: ++

Assets/MirageXR/DataModel/DataModel.cs
cawke: I'm not quite sure yet how it will work. I think we'll clarify this at the time of implementation.

cawke: ++

Assets/MirageXR/DataModel/DataModel.cs
cawke: ++

Assets/MirageXR/DataModel/DataModel.cs
cawke: we discussed with misha that not all triggers can be used on this or that content. this array shows which triggers can be applied to this content

Assets/MirageXR/DataModel/DataModel.cs
cawke: ++

Assets/MirageXR/DataModel/DataModel.cs
cawke: ++

Assets/MirageXR/DataModel/DataModel.cs
cawke: addressable cannot be part of the data model. in this case it (FileName) would be id addressable

Assets/MirageXR/DataModel/DataModel.cs
cawke: Now we store them as addressables. We just need an id to access them.

Assets/MirageXR/DataModel/DataModel.cs
fwild: I don't think that works, because what to do with downloaded units? The local guid assigned by a different user is not resolvable. We are changing login now to SSO on the identity server keycloak - we can try to bring the guid in line with the server uuid, but I think it is easier to resolve if this is the uuid, not the (local!) guid.

Assets/MirageXR/DataModel/DataModel.cs
fwild: Ok

Assets/MirageXR/DataModel/DataModel.cs
fwild: Go it

fwild (changes requested):
Almost all resolved - just the user uuid?

cawke:
> Almost all resolved - just the user uuid?

They write on the internet that "GUID (Globally Unique Identifier) is equivalent to a UUID". 
in json file guid is serialized as string, on server string is deserialized as uuid and vice versa.

cawke:
> > Almost all resolved - just the user uuid?
> 
> They write on the internet that "GUID (Globally Unique Identifier) is equivalent to a UUID". in json file guid is serialized as string, on server string is deserialized as uuid and vice versa.

@fwild ^^^

Assets/MirageXR/DataModel/DataModel.cs
```diff
@@ -0,0 +1,319 @@
...
+    public class Comment : DataModel
+    {
+        [JsonProperty] public User User { get; set; }
+        [JsonProperty] public string Text { get; set; }
+        [JsonProperty] public List<File>? Attachment { get; set; }
+    }
+
+
+    [Serializable]
+    public class RequiredToolsPartsMaterial
```
fwild: should be `RequiredToolsPartsMaterials` (plural at the end)

fwild (changes requested):
spotted a missing plural 's'

fwild:
> > Almost all resolved - just the user uuid?
> 
> They write on the internet that "GUID (Globally Unique Identifier) is equivalent to a UUID". in json file guid is serialized as string, on server string is deserialized as uuid and vice versa.

Fine - as long as the unique ID comes ultimately from the keycloak server.

Assets/MirageXR/DataModel/DataModel.cs
cawke: done

Assets/MirageXR/DataModel/DataModel.cs
```diff
@@ -0,0 +1,319 @@
...
+    {
+        [JsonProperty] public Guid Id { get; set; }
+        [JsonProperty] public DateTime CreationDate { get; set; }
+        [JsonProperty] public string Version { get; set; }
+    }
+
+    [Serializable]
+    public class Activity : DataModel
+    {
+        [JsonProperty] public User Creator { get; set; }
```
BenediktHensen, line 21: I think it would make sense to have multiple creators here once we have a collaborative editor. Or at least also to add further contributors apart from one main creator as a separate field.

Assets/MirageXR/DataModel/DataModel.cs
```diff
@@ -0,0 +1,319 @@
...
+using UnityEngine;
+
+namespace DataModel
+{
+    [Serializable]
+    public class DataModel
+    {
+        [JsonProperty] public Guid Id { get; set; }
+        [JsonProperty] public DateTime CreationDate { get; set; }
+        [JsonProperty] public string Version { get; set; }
```
BenediktHensen, line 15: Is this the file version or the version of MirageXR that was used to create this?
If it is the file version, maybe an integer would suffice that we can increment with each upload / major edit?

Assets/MirageXR/DataModel/DataModel.cs
```diff
@@ -0,0 +1,319 @@
...
+        [JsonProperty] public List<Guid>? StepIds { get; set; }
+        [JsonProperty] public List<HierarchyItem>? Item { get; set; }
+        [JsonProperty] public string Title { get; set; }
+        [JsonProperty] public string Description { get; set; }
+    }
+
+    [Serializable]
+    public class ActivityStep : DataModel
+    {
+        [JsonProperty] public uint Order { get; set; } //or it can be based on Activity.Steps?
```
BenediktHensen: Yes, the Order property is probably redundant data which should be derived from the Activity.Steps. Using the Order property could also lead to trouble as there can now be multiple ActivitySteps, e.g., with Order set to 0.

Assets/MirageXR/DataModel/DataModel.cs
```diff
@@ -0,0 +1,319 @@
...
+    public class Content<T> : DataModel where T : ContentData
+    {
+        [JsonProperty] public ContentType Type { get; set; }
+        [JsonProperty] public string Name { get; set; }
+        [JsonProperty] public Location Location { get; set; }
+        [JsonProperty] public T ContentData { get; set; }
+    }
+
+    [Serializable]
+    public class Trigger    //todo
```
BenediktHensen, line 142: It could be beneficial to break this class down into subclasses that inherit from Trigger. At the moment, it is a general purpose class that contains configurations for sensors, LLMs and others, so having a clearer separation could help giving the trigger clearer purpose.

Assets/MirageXR/DataModel/DataModel.cs
```diff
@@ -0,0 +1,319 @@
...
+    [Serializable]
+    public class GlyphContentData : ContentData
+    {
+        [JsonProperty] public string FileName { get; set; }
+    }
+
+    [Serializable]
+    public class ImageContentData : ContentData
+    {
+        [JsonProperty] public string Text { get; set; }
```
BenediktHensen, line 275: What would the Text be for? Is it a caption?

Assets/MirageXR/DataModel/DataModel.cs
fwild: was intended for MirageXR version, so we can create converters in the future...

Assets/MirageXR/DataModel/DataModel.cs
fwild: Alt text, yes.

fwild: currently displayed as caption

fwild:
Looks like only some small cleanup is needed and we are good to go :)

Assets/MirageXR/DataModel/DataModel.cs
```diff
@@ -0,0 +1,319 @@
...
+        [JsonProperty] public string Name { get; set; }
+        [JsonProperty] public Location Location { get; set; }
+        [JsonProperty] public T ContentData { get; set; }
+    }
+
+    [Serializable]
+    public class Trigger    //todo
+    {
+        [JsonProperty] public TriggerType Type { get; set; }
+        [JsonProperty] public Guid StepId { get; set; } // step.id
```
fwild: `JumpToStepId`? And it can also be NULL (to allow OutputEvents to be generated WITHOUT leaving the step)

Assets/MirageXR/DataModel/DataModel.cs
```diff
@@ -0,0 +1,319 @@
...
+        [JsonProperty] public string Event { get; set; } // string needs to be eval’d
```
fwild: Maybe we need to separate between:
* `InputEvent` (e.g. "onIncorrectPlacement" from pick and place: this trigger listens to this event, and if it occurs, and if it is an Event of type trigger, then it triggers the trigger) 
* and `OutputEvent` (an event to be thrown, when the trigger conditions are fulfilled - to inform all augmentations about trigger, e.g. "show"/"hide"/"animate augmentation=5")

Assets/MirageXR/DataModel/DataModel.cs
```diff
@@ -0,0 +1,319 @@
...
+
+    [Serializable]
+    public enum TriggerType     //todo
+    {
+        Unknown = -1,
+        Click = 1,
+        Sensor = 2,
+        Detect = 3,
+        Event = 4 // maybe? From improved pick&place, that uses events to advance?
+        //todo
```
fwild, line 168: add `Proximity = 5`

Assets/MirageXR/DataModel/DataModel.cs
fwild: * and `OutputEventParameter` + `OutputEventParameterValue`

fwild: * Mikhail and I think that the trigger should also have an optional `ContentId` so that it is linked with its augmentation

Assets/MirageXR/DataModel/DataModel.cs
cawke: yes, it sounds good, but for now I have a poor idea of ​​all the triggers to describe them. I think as we implement them we will expand this model.

Assets/MirageXR/DataModel/DataModel.cs
cawke: ++

Assets/MirageXR/DataModel/DataModel.cs
cawke: ++

Assets/MirageXR/DataModel/DataModel.cs
cawke: ++

Assets/MirageXR/DataModel/DataModel.cs
fwild: ok, so let's leave it at that for now. Marking as resolved.

fwild (approved):
All comments are resolved in my view, there fore approving.

PR #2060, opened 2024-08-15
Create macos_develop_workflow.yml

fwild:
first test


fwild:
@cawke - this is really just using the right runner type keyword - using a github provided free runner. Wondering if we can get this to run easily?

Issue #2051, opened 2024-08-12
Update the Speech Settings of the Virtual Instructor based on the design review


fominykh:
The design review has been done in this issue: https://github.com/WEKIT-ECS/MIRAGE-XR/issues/2023
Multiple layout and font issues identified.
@Alvareon can you check how the current UI of the Speech Settings is done? And after that talk to @AKissMail ?

Alvareon:
@fominykh @AKissMail I've updated the settings panel according to the Figma design. I've moved the repeating elements into a separate prefab and updated the layout: PR https://github.com/WEKIT-ECS/MIRAGE-XR/pull/2063


Additionally, the following screens need to be fixed: ContextPrompt, Voice, Models, and Language. Please also check each screen at different resolutions using Simulator mode in Unity."

<img width="453" alt="Screenshot 2024-08-21 at 00 28 24" src="https://github.com/user-attachments/assets/e27a0622-bee0-487d-bd3c-6fc5386442ca"><img width="444" alt="Screenshot 2024-08-21 at 00 21 34" src="https://github.com/user-attachments/assets/05a55889-c65f-4398-942c-ce01492f859b">

AKissmehl:
@fominykh Can we close that one?

fominykh:
@AKissMail I didn't see it in a build, but we can close this one. I will do another review later if needed and will open a new issue for it.

PR #2063, opened 2024-08-20
updated speech settings panel

Alvareon:
https://github.com/WEKIT-ECS/MIRAGE-XR/issues/2051
I've updated the settings panel according to the Figma design. I've moved the repeating elements into a separate prefab and updated the layout.

| Screenshot 2024-08-21 at 00 21 34 | Screenshot 2024-08-21 at 00 28 24|
| ------------- | ------------- |

### DIFF
--- a/Assets/MirageXR/ContentTypes/VirtualInstructor/Editors/Mobile/Prefabs/SettingsListItem.prefab
+++ b/Assets/MirageXR/ContentTypes/VirtualInstructor/Editors/Mobile/Prefabs/SettingsListItem.prefab
@@ -1,0 +1,743 @@
+%YAML 1.1
+%TAG !u! tag:unity3d.com,2011:
+--- !u!1 &198465672100755037
+GameObject:
+  m_ObjectHideFlags: 0
+  m_CorrespondingSourceObject: {fileID: 0}
+  m_PrefabInstance: {fileID: 0}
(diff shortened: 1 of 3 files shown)


AKissmehl:
Mmm, that was probably unfortunate communication, but was that not my task? At least I worked on it and also have a PR ready in a minute. I would suggest taking the best form of both of us, I did not use prefabs, but it makes total sense. I did notice additional 2 issues in the PR. 

1. The spacing looks a bit weird in the setting scenes now. 
Screenshot 2024-08-21 at 12 27 30
Screenshot 2024-08-21 at 12 28 08

2. The Model tile is missing the on click event.

When I was working on it, I noticed that there is not a model set by default. I added a start function that sets the first models if it is null. Moreover, I move the button to the tile, instant to have it just on the arrow. 

I am short of to-dos currently, should I merge that together and fix the spacing in the sub scenes?

Alvareon:
@AKissMail I haven't updated the sound and language screens (they also need to be updated). You are free to use my changes in your PR. I also recommend increasing the distance from the top edge of the screen to the header so that the phone notch doesn't cover the UI elements.

Alvareon:
pr closed (see https://github.com/WEKIT-ECS/MIRAGE-XR/pull/2065)

PR #2074, opened 2024-08-29
OIDC login

### LINKED ISSUE #2064: OIDC login
Add login support for OIDC via identity.wekit-ecs.com (KeyCloak)

fwild:
OIDC with https://identity.wekit-ecs.com works, stores the token locally and retrieves it instead of logging in again when opening the app next time. The logout button in profile also properly logs out and removes the stored token from config.info. User Info is retrieved from the endpoint.

Resolves #2064


fwild:
https://github.com/user-attachments/assets/024fb5c6-8269-41eb-8fb4-1412b51de334

fwild:
https://github.com/user-attachments/assets/c90c7460-4ce7-4c8d-b5ab-498ee77c86c6

AlexanderNeumann:
The userinfo should be accessible if the authentication expects the `openid` scope. Regarding the last screen recording it seems to work, right?

fwild:
Changes implemented as discussed, @cawke

cawke (approved):

PR #2085, opened 2024-09-18
Andreas/2080 event trigger for the virtual instructor

### LINKED ISSUE #2080: Event Trigger for the Virtual Instructor
Refactoring of the VirtualInstructorManager into a VirtualInstructorOrchestrator. Therefore, the business logic of the VirtualInstructorManager has to be adapted. Mainly, the AskClosestInstructor function will be split into multiple functions:

1.     DetermineVirtualInstructor Should determine the relevant Instructor in space. A relevant Instructor is considered to be within the user's field of view and closer than other instructors.

3.     AskClosestInstructor(AudioClip) Takes an Audio Clip and formulates an answer based on the question. The return value contains the answer from the LLM/RAG model as a String and Audio and is based on the result of the LLM. This function uses DetermineVirtualInstructor to choose a VI.
 
5.     AskClosestInstructor(String) Takes a String and formulates an answer based on that String and the content of the Queue. The return value contains the answer from the LLM/RAG model as a String and Audio and is based on the result of the LLM.
 
7.     AddToQueueOfClosestInstructor(String) Adds the relevant information as a String to the Queue of the VI. This can later be used as contextual information for a prompt.

Functions 3 and 4 will be listening to events that follow the pattern of the gaze trigger, but it has to be discussed what the final implementation of that looks like.

AKissmehl:
This PR refactors the Virtual Instructor Manager into an orchestrator and introduces new methods. This can be accessed via the root object to interact with one or multiple VI's. It is now possible to add strings to the next question of a user, to create a statement by the Virtual Instructor without the use of an LLM/RAG model, and to send a question to a VI based on a string. Additionally, the code itself was refactored to make it more maintainable.

One open question is the structure of the events. Currently, we are using the root object, but that may not be the ideal solution. In my view, an event manager where events can be subscribed/unsubscribed would be better, but this would require more time, and based on the last meeting, I understand that we need to move forward.


fwild (approved):
Only comment: shouldn't "VirtualInstructorRecoder.cs" be "VirtualInstructorRecorder.cs"?

PR #2086, opened 2024-09-18
Collaborative Avatars

BenediktHensen:
This pull request adds collaboration and users visualized as avatars using Photon Fusion and ReadyPlayerMe.

It updates to Unity 2022.3.33f1 since a Photon feature required this update during the initial package import.

For developers without access to Photon Fusion, the project should work as normal - code is deactivated using the FUSION2 directive. This directive gets automatically enabled if the package is imported, so sharing features should automatically start working after the import. In the scene, developers without Photon Fusion will see a broken prefab - I will check if there is an alternative with a runtime initialization instead.

Updating this to the latest develop state was difficult as some GUIDs in Unity's files where conflicting and broke the XR configuration. I restored the configuration so that all errors disappear but it should be checked if this version runs on the hardware.


fwild (commented):
we should switch to MirageXR assemblydef by name, not by GUID;
yes, the start scene gameobject (ConnectionManager) needs to be instantiated programmatically, otherwise we have a broken project if anyone does not have the right libraries imported;
otherwise it seemed to have worked for me;
after package import, there were some issues flag with meta files (but I don't think they are in here - must have been leftover from a previous merge - they did contain merge conflicts still)

fwild (changes requested):
these changes need to be adressed before we can merge

fwild:
And I propose to update to Unity 2022.3.46f1 to bring it inline with the AppleVision Pro branches. This did not affect anything for me, though it may be worth checking out this branch's package dependencies to compare: https://github.com/WEKIT-ECS/MIRAGE-XR/blob/fwild/visionos-build-test3-readyplayerme/Packages/packages-lock.json

BenediktHensen:
The requested changes are now added:

- Added a manager which adds the Fusion prefab to the scene at runtime if Fusion is installed
- Updated to Unity 2022.3.47f1
- switched MirageXR assembly definition to reference by name

BenediktHensen:
The latest version fixes the PhotonAddon compiler errors - it removes Photon Addons and moves only the necessary components into the own Networking scripts folder. (requires the newest Photon config zip)

fwild (approved):

cawke (approved):

PR #2094, opened 2024-09-24
Compile fixes for the Microsoft HoloLens

BenediktHensen:
This pull request fixes compiler errors that happened on the UWP build target:

- There were several pieces of code that still called functions, e.g., of the EventManager, that were moved into lib-lee. This caused compiler errors as the original function did not exit anymore. I have fixed these errors and reconnected them with their counterparts in lib-lee.
- I removed the Microsoft.MixedReality.DotNetWinRT package as it caused a strange compiler error in the MRTK where a Windows namespace was searched but not found. `Assets\MRTK\Core\Utilities\WindowsApiChecker.cs(10,17): error CS0234: The type or namespace name 'Windows' does not exist in the namespace 'Microsoft' (are you missing an assembly reference?)` This fix of removing the DotNetWinRT package gets rid of the compile error but it should be tested whether this package was necessary for any core functionality.

(Also contains the update to Unity 2022.3.47 since we use that version on many other branches now)


fwild (commented):
the dotwinRT was needed to build, I think? Did you try to build?

fwild:
Thanks for confirming, @BenediktHensen, that WinRT is not needed.

fwild (approved):

PR #2101, opened 2024-10-04
VisionOS support

Herman1412:
Adds visionOS mixed reality support using polyspatial.

Note to developers:
The project requires a educational or pro license to open due to the polyspatial packages. 
Vuforia is not supported on visionOS, and visionOS is not excluded as a platform in the Vuforia scripts assembly. When building, visionOS will have to be manually added to the assembly definition found at: _Packages/com.ptc.vuforia.engine/Vuforia/Scripts/VuforiaScripts.asmdef_.

More info on visionOS integration can be found on the wiki: [UI button interaction for vision os](https://github.com/WEKIT-ECS/MIRAGE-XR/wiki/UI-button-interaction-for-vision-os)


fwild:
I propose we either remove the Assets/Samples/ completely or we put it into the .gitignore? I believe you cloned the XR rig anyways, so it should be possible to remove them? They just blow up the project, and should not be needed ...

fwild (approved):

PR #2104, opened 2024-10-15
 Add virtual instructor

AKissmehl:
This PR introduces several new prefabs needed for the scenes of the virtual instructor editor.


fwild (approved):
Looks good to me, but would be good if Sergey glimpses over this as well!

AKissmehl:
I indeed ask him rn. There is also one small thing missing where I need a references or tutorial. So there will be an update anyway soon.

Alvareon:
@AKissMail if you plan to implement these panels as pop-ups, then these panels need to be removed from the main prefab: 

Screenshot 2024-10-15 192044

Alvareon:
VISettings.prefab height does not match the height of the panel in Figma (251 vs 220)

PR #2105, opened 2024-10-15
new activity manager

cawke:


cawke:
https://github.com/WEKIT-ECS/lib-lee/pull/13

PR #2120, opened 2024-10-20
step view and intsructor

cawke:
@AKissMail I did not find in your script the data editor for the model that is created/changed. Add them and fill the model with them in the OnAccept method of the AddEditVirtualInstructor class.
image

once you do this check the instructor's work and delete these lines (InstructorContentView.InitializeContentAsync(Content<InstructorContentData> content)) . I don't know how it should work
image


cawke:
https://github.com/WEKIT-ECS/MIRAGE-XR/pull/2123

PR #2146, opened 2024-11-07
Andreas/2134 fix virtual instructor on spatial UI

AKissmehl:
Update the port in the config


Herman1412:
The two red buttons are not visible when adding an instructor

AKissmehl:
Okay, I will take a look! — Thanks for testing!

Assets/Resources/AI_Server.txt
```diff
@@ -1,3 +1,3 @@
-AI_API_URL=http://91.107.198.4:8000/
+AI_API_URL=http://127.0.0.1:8001
```
fwild, line 1: This would have to be a public URL, maybe we can also switch to the domain now? repository.wekit-ecs.com:8000 should be the public master release, and :8001 should be for develop (currently isn't)?

fwild (changes requested):
the pipeline changes are ok, but have not been tested....? And the localhost url needs to be changed...

Issue #2124, opened 2024-10-23
Spatial Keyboard


fwild:
We need to integrate the spatial keyboard for spatialUI.

PR #2147, opened 2024-11-10
Herman1412/spatial keyboard

Herman1412:
Keyboard can be used with tmp input fields and will spawn in front of the player
Can be tested using the input field on main screen.

Close issue #2124


Assets/Samples/XR Interaction Toolkit/3.0.5/Spatial Keyboard/Scripts/GlobalNonNativeKeyboard.cs
fwild, line 1: Is this file needed? If it is modified from the samples, we should move it into our own Assets/MirageXR/... folder.

Assets/XRI/Settings/Resources/XRDeviceSimulatorSettings.asset
fwild, line 1: Can we remove this change from the PR? Otherwise we will be rolling back and forth between different developers' needs?

Assets/Resources/fonts/LiberationSansSDF.asset
fwild, line 1: fine to leave it in, but this also creates a cyclic change - between cyrillic and other locales, so best to 'git restore' the file and not committing it in the future.

fwild (approved):

fwild (approved):

PR #2150, opened 2024-11-11
Hover effect on buttons

Herman1412:
Check hover on buttons when selecting augmentation type and hover on elements on profile screen.


fwild (approved):
Shows as 'draft' for me, so wasn't sure I should review already?

Herman1412:
Left it as draft in case there are more buttons that should be changed, did not want it merged right away until I hear back from Mikhail

PR #2160, opened 2024-11-14
Avatar Blinking Animation

BenediktHensen:
This pull request makes the avatar blink every ten seconds using ReadyPlayerMe's blend shapes.


fwild (approved):
Though adults have about 4sec blink frequency, not 10 - maybe change?

BenediktHensen:
The blinking intervals of adults are normally between 2 and 10 seconds but also depend greatly on other factors such as concentration, etc. However, recommendations by animators are to choose at least the top boundary of that range since fully realistic values can be too prominent and distracting. It is supposed to be a subtle effect that the user only notices subconciously to give the impression that the avatars are more life-like but should not be actively perceived. I tested the different frequencies and lower intervals make the blinking a lot more noticable up to the point where it becomes the main focus point in interactions.

fwild:
convinced :)

Issue #2137, opened 2024-11-06
lip-sync for avatars


fwild:
Hook up the lip sync for the avatars

Issue #2038, opened 2024-07-10
Collaboration: C7: Lipsync for avatars


fwild:
Add lipsync support for readyplayer me avatars - probably with Oculus lipsync:
https://docs.readyplayer.me/ready-player-me/api-reference/avatars/morph-targets/oculus-ovr-libsync

PR #2161, opened 2024-11-14
Lip Syncing

BenediktHensen:
This pull request implements #2038 and #2137.

It adds lip sync via the Oculus Lip Sync library.

The lip sync components are initialized once an avatar is successfully loaded. I adjusted the ReadyPlayerMe download settings so that it now contains the blendshapes for the necessary phonemes. Once the component is created, the library accesses them and takes control of these blendshapes. The functionality of the library had to be adjusted slightly since Unity switched from blendshape values between 0 and 100 to 0 and 1.

To connect the lip syncing, I refactored the prefab structure for the avatar and the voice chat components. The recorder now exists from the beginning as part of the connection manager object (but only starts recording once in a collaborative session) and the speaker is now a direct part of the avatar.

During the refactoring, I also improved the voice quality. The voice is now louder since the speaker is positioned on the avatar's mouth. Moreover, the spatial position of the sound source can be perceived since spatial blending is enabled - if the avatar is on the left side of the user, the voice also appears to be coming from the left.


fwild:
How to we roll back if the new OVR libraries pose a problem on visionOS? I did not see any asmdef visionOS specific statements, could you check if this is in the asmdef (or if there is an exclude where it might be needed)?

BenediktHensen:
We can always roll back either via Git or by deleting the library and the used components again. We need to confirm visionOS functionality ourselves but since I don't have the device, I need your feedback on this. The library is part of a Meta Audio Spatializer suite which is not maintained anymore, so it is not adjusted to the newest devices - we would need to set any defines or make adjustments for visionOS ourselves.

PR #2191, opened 2024-11-21
Step Interaction fix

cawke:
step interaction has been fixed
model list item mask has been fixed
interaction samples has been added


cawke:
@fwild you changed the mask color to transparent. was it just a mistake or was it intended?

image
image
image

fwild (approved):

PR #2256, opened 2025-01-21
main scene fix

cawke:
dont merge. 

@fwild first need to test this fix on vision os. not sure how it will work.

fixes for event systems. allows testing UI with mouse (without these changes clicking doesn't work at least for me and @AKissMail) and interacting with objects using controller simulators.

I thought that the problem with event system was only mine, that's why I didn't commit them earlier.

image

### DIFF
--- a/Assets/MirageXR/Scenes/Start.unity
+++ b/Assets/MirageXR/Scenes/Start.unity
@@ -360,6 +360,100 @@ Transform:
   m_Children: []
   m_Father: {fileID: 930107308}
   m_LocalEulerAnglesHint: {x: 0, y: 0, z: 0}
+--- !u!1 &71129122
+GameObject:
+  m_ObjectHideFlags: 0
+  m_CorrespondingSourceObject: {fileID: 0}
+  m_PrefabInstance: {fileID: 0}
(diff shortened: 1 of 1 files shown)


fwild:
I have tested the build from Herman now as well, this is what I found:
* it seems the volume camera setup has changed, and MirageXR now starts non-exclusive (= you can see the other windows like TestFlight that I had open still, while MirageXR is open). That's per see not a totally bad thing, but it needs a discussion, whether we want that - should not affect the interaction use in any case.
* I can see dark (=cutout) virtual hands, but at the same time the VisionOS cutouts around my hands are not visible. I think this is because we switched to the other Volume Camera setting? Or is this because we activated hands visualisation? They are clearly cut off hands, whereas the native apple support for upper limb visibility would not only cut out the "glove" shape, but the full arm
* interaction works
* interaction did not work on the toggle button of the collaboration window and not on the view/edit toggle, so I could not see how the virtual hands cutout works together with the avatar visualisation (I guess it does not fit well, but could not test it)

So: overall, it seems to work, but we should discuss the virtual hands and the volume camera setting, and we should make sure that the new events (?) are assigned to every UI element.

That's my five pence.

PR #2272, opened 2025-01-29
Sprites Fix

cawke:


fwild (approved):

fwild:
we should subsequently do an updated release for 2.7

PR #2280, opened 2025-02-19
Auto vision setup (experiment)

cawke:
test on vision os first.

in this branch XRTouchSpaceInteractor will be added automatically when a prefab appears in the scene.

goal is not to clutter prefabs with highly specialized components (XRTouchSpaceInteractor for each interactive component) (not sure if XRTouchSpaceInteractor is needed for vision os)

the branch uses a hack in the form of Harmony. with its help, access to the event that some prefab was instantiated appears.


fwild:
I asked Herman to do a build to test

fwild (approved):

PR #2281, opened 2025-02-20
Release 2.7 to Develop

cawke:


ProjectSettings/ProjectSettings.asset
```diff
@@ -267,7 +273,7 @@ PlayerSettings:
   AndroidTargetDevices: 0
   AndroidSplashScreenScale: 0
   androidSplashScreen: {fileID: 0}
-  AndroidKeystoreName: '{inproject}: Keystore/miragexr.keystore'
```
fwild, line 270: might have to be removed? I think we did not keep it in develop, only set it for releases, right?

ProjectSettings/ProjectSettings.asset
```diff
@@ -1035,4 +1051,4 @@ PlayerSettings:
   hmiLoadingImage: {fileID: 0}
   platformRequiresReadableAssets: 0
   virtualTexturingSupportEnabled: 0
-  insecureHttpOption: 0
+  insecureHttpOption: 2
```
fwild, line 1054: Hm. Please cross check - as some of our servers do not have ssl active? Is this the option that allows or forbids insecure http?

fwild (changes requested):
and please check after that we are not leaving something in Assets/DevDunkStudio/ - as we moved this directory out to MirageXR-conciliator...

fwild:
also there seem to be some merge conflicts in the ISO warning signs meta files...

ProjectSettings/ProjectSettings.asset
cawke: image

ProjectSettings/ProjectSettings.asset
cawke: Yes, we don't store the key in the repository and it is used only for the release

cawke:
https://github.com/WEKIT-ECS/MIRAGE-XR/pull/2283

Issue #2259, opened 2025-01-23
Spatial hyperlinks


fwild:
Create the markdown style functionality:
- parse the description text for square brackets "[" and "]"
- change color/add background colour as button, so that the link text becomes clickable
- call John's bezier splines function (or prefab) with target location+orientation as parameter (which then activates a bezier curve to come from an inch in front of the pointer finger, leading directly to the target location)

![Image](https://github.com/user-attachments/assets/658f0d78-af2e-43d3-aeb9-794032b092d3)

PR #2285, opened 2025-03-03
Spatial hyperlinks

Alvareon:
https://github.com/WEKIT-ECS/MIRAGE-XR/issues/2259

Description:

This pull request introduces functionality for spatial hyperlinks and Bezier curves, enhancing the description field with parsing capabilities and spatial coordinate management. Key features include:

- Spatial Hyperlinks: Added support for creating spatial hyperlinks using square brackets (e.g., [ [link]]()). These hyperlinks are now visually and spatially integrated into the UI, allowing users to interact with 3D objects in the scene.
- Bezier Curves: Implemented Bezier curve generation to connect hyperlinks with their corresponding 3D objects, providing smooth visual paths in the spatial environment.
- Description Field Parsing: Developed parsing logic to process the description field’s markup, identifying and handling hyperlinks and formatting within the text.
- Spatial Coordinate Saving: Enabled the storage and management of spatial coordinates for hyperlinks, ensuring their positions are preserved and updated accurately in the space.


Alvareon:
Before merging, it is necessary to test on a real device

Assets/MirageXR/Player/Scripts/Spatial/InfoScreenSpatialView.cs
```diff
@@ -43,9 +62,319 @@ public void Initialize(ActivityStep step)
             UpdateView(step); 
         }
         
+        private void Update()
+        {
```
fwild, line 71: This will not work on the glasses - this should be a different, additional event?

Assets/MirageXR/Player/Scripts/Spatial/InfoScreenSpatialView.cs
```diff
@@ -43,9 +62,319 @@ public void Initialize(ActivityStep step)
...
+            {
+                Destroy(spline.gameObject);
+                _splineInstances.Remove(linkId);
+            }
+            _previousPositions.Remove(linkId);
+        }
+        
+        private void HandleClick()
+        {
+            var mousePosition = Input.mousePosition;
```
fwild, line 148: This will not work on the smart glasses - needs to be a Spatial Pointer Device interactionPosition or some such, see here: https://docs.unity3d.com/Packages/com.unity.polyspatial.visionos@1.3/manual/PolySpatialInput.html

fwild: There are good examples on the page. If not via direct scripting (which I think is better), then HandClick() could also be triggered via the XRTouchSpaceInteractor, but I think this is more complicated, and @cawke is currently updating the use of XRTouchSpaceInteractor?

fwild (changes requested):
I think the click interaction will not work on the smart glasses - would need interaction from https://docs.unity3d.com/Packages/com.unity.polyspatial.visionos@1.3/manual/PolySpatialInput.html

Assets/MirageXR/Player/Scripts/Spatial/InfoScreenSpatialView.cs
Alvareon: fixed

Assets/MirageXR/Player/Scripts/Spatial/InfoScreenSpatialView.cs
Alvareon: fixed

fwild (approved):

PR #2293, opened 2025-03-04
fixed grabber on ViewSpatial home screen

### LINKED ISSUE #2292: grabber bar on main menu broken
Grabber bar does not work on the main dialogue.

fwild:
Resolves #2292


fwild:
Setting it exactly to same values as on profile - please nod off, no need to check deeply, we need to check in app anyways.

Herman1412 (approved):
works on sim

PR #2299, opened 2025-03-04
Fwild/2296 labelfixes

fwild:
some improvements:
- layerorder for label background (experiment)
- left button on label removed (no function)
- materials for line and sphere updated to have shadow and visibility (line previously not showing)
- attach rotation set to false (hopefully fixing the jerk rotation prob with movement)


fwild:
did not work - canning.

PR #2305, opened 2025-03-06
updated saving position for spatial anchor

Alvareon:


fwild (changes requested):
I think this is not going to do the job - we need to capture the onRelease event of the drag and drop of the anchor instead. Otherwise collaboration will not be able to share, and if, e.g., deeplinking opens another activity etc., the position will not be saved).

fwild:
I think the SaveHyperLinkPosition should go into line 92 of SpatialHyperlinkObjectView.cs:
```
   private void OnManipulationEnded() { }
```
and think the SaveHyperLinkPosition would be better in there as well - separation of concerns, making sure that the data storage is with the view (or rather a separate business logic script that is called by the view? to make sure we can reuse the same method in the mobile views).

fwild (approved):

PR #2317, opened 2025-03-20
control fix

cawke:


Assets/MirageXR/Managers/ViewManager/ViewManager.cs
```diff
@@ -77,8 +85,23 @@ private void OnActivityLoaded(Activity activity)
             }
         }
 
+        private void CreateCamera()
+        {
+#if VISION_OS && !UNITY_EDITOR
```
fwild: might cause trouble with play to device - need to check, I think, or at least check after!

Assets/MirageXR/Player/Scripts/RootObject.cs
```diff
@@ -102,9 +97,9 @@ private async Task Initialization() // TODO: create base Manager class
 
 			try
 			{
-#if POLYSPATIAL_SDK_AVAILABLE && VISION_OS
-                InstantiateExtensions.Initialize();
-#endif
+// #if POLYSPATIAL_SDK_AVAILABLE && VISION_OS
+//                 InstantiateExtensions.Initialize();
```
fwild, line 101: what were those again?

Assets/MirageXR/Player/Scripts/RootObject.cs
```diff
@@ -129,7 +122,6 @@ private async Task Initialization() // TODO: create base Manager class
 				_floorManager ??= new GameObject("FloorManagerWrapper").AddComponent<FloorManagerWrapper>();
 				_floorManagerWithRaycastFallback ??= new GameObject("FloorManagerWithRaycastFallback").AddComponent<FloorManagerWithFallback>();
 				_pointCloudManager ??= new GameObject("PointCloudManager").AddComponent<PointCloudManager>();
-				_volumeCameraManager ??= new GameObject("VolumeCameraManager").AddComponent<VolumeCameraManager>();
```
fwild, line 132: so we don't use this anymore? don't need it?

Assets/MirageXR/View/Content/InstructorContentView.cs
```diff
@@ -1,13 +1,11 @@
 using Cysharp.Threading.Tasks;
-using GLTFast.Schema;
 using i5.Toolkit.Core.VerboseLogging;
 using LearningExperienceEngine.DataModel;
-using ReadyPlayerMe.Core;
```
fwild, line 5: surprised it is not needed here - as we now have built in readyplayerme support for the VI characters...?

fwild (commented):
I think this looks good, let's wait for Herman to do a build before merge - and I left some comments where I was not fully sure...

Assets/MirageXR/Player/Scripts/RootObject.cs
cawke: here we initialized a hack to receive an event that an object was instantiated. it was necessary to automatically add `XRTouchSpace`. As far as I understand, we don't need `XRTouchSpace` or we only need it in one instance.

Assets/MirageXR/View/Content/InstructorContentView.cs
cawke: this means that this namespace is used in another component.

Assets/MirageXR/Player/Scripts/RootObject.cs
cawke: I add it differently.

cawke:
```
        private void CreateCamera()
        {
#if VISION_OS && !UNITY_EDITOR
            var prefab = _assetBundleManager.GetCamera(CameraType.VisionOS);
#else
            var prefab = _assetBundleManager.GetCamera(CameraType.OpenXR);
#endif
            _cameraView = Object.Instantiate(prefab);
            _camera = _cameraView.GetComponentInChildren<Camera>();
        }
```

PR #2321, opened 2025-03-20
update project to unity 6

Herman1412:
Update to unity 6000.0.40f1 in order to use polyspatial 2.2.4


fwild (approved):
I have read over this, and I think it is fine, but am awaiting Alex's second review

cawke (approved):

PR #2333, opened 2025-03-27
photon voice fix

cawke:


cawke:
need to check on the device

fwild (approved):

PR #2355, opened 2025-04-30
Vision os fix shaders

Herman1412:
The polyspatial version has been downgraded to v2.0.4 and this fix has been added to bring back the shaders: https://github.com/HoloLabInc/glTFast.git#polyspatial-urp-transparent


fwild (approved):
@cawke please also give quick glimpse, also to make sure it does not affect mobile (shouldn't)...

cawke (approved):

PR #2386, opened 2025-06-04
vision os camera offset has been added

cawke:
Need to be tested first


fwild:
tested, works, merging.

PR #2390, opened 2025-06-05
temp fix for DeepLink

cawke:
I'm not sure if this is necessary right now, but it might be useful for a build on vision os.


fwild (changes requested):
Please commit this directly to i5, so we can just upgrade the package after a next release: https://github.com/rwth-acis/i5-Toolkit-for-Unity/blob/upm/Tests/Editor/Deep%20Link%20Service/DeepLinkTestDefinition.cs

fwild:
Yes, currently the login works - as you established, we /think/ that it was the setting to reduce code stripping to minimum, but there is a small chance that it might have been something else?

cawke:
> Yes, currently the login works - as you established, we /think/ that it was the setting to reduce code stripping to minimum, but there is a small chance that it might have been something else?

I've checked that the reason is Stripping + System.Reflection

PR #2406, opened 2025-06-20
permission manager

cawke:
https://github.com/WEKIT-ECS/lib-lee/pull/53


fwild (approved):
Please double check if the Packages/com.readyplayerme.core@ really needs to be included here? I thought we are using this package vanilla from the package manager?

PR #2418, opened 2025-07-30
Add activity list fetch and copy/paste functionality

cawke:
https://github.com/WEKIT-ECS/lib-lee/pull/55

image

image


fwild:
If this is the branch tested in 2.8.4 (build 2), then it fails to start the app on my visionPro, I had to roll back to 2.8.4 (build 1)

Issue #2437, opened 2025-08-15
refactor VirtualInstructor to remove _history and use thread_id instead (and pass on to AI manager in lib-lee)


fwild:
* add an VirtualInstructor.initialize (provide: name, model, instructions) -- this method then calls an AImanager.initialize to receive the thread_id and assistant_id 
* update AskVirtualInstructorAudio (and AskVirtualInstructorString) to include thread_id and assistant_id
* update lib-lee/ArtificialIntelligenceManager to include thread_id  and assistant_id* 
* remove {_history} (and context composition for the history), when using the thread_id / assistant_id, remove update history

See lxr implementation:
* https://github.com/WEKIT-ECS/lxr/blob/development/llm/vendors/gpt_open_ai.py
* https://github.com/WEKIT-ECS/lxr/blob/development/llm/views/initialize_agent_http_view.py
 
Here the MirageXR code:
* https://github.com/WEKIT-ECS/MIRAGE-XR/blob/develop/Assets/MirageXR/ContentTypes/VirtualInstructor/VirtualInstructor.cs
* plus the according changes in lib-lee

AKissMail:
Also, we need to pick the thread_id out of the response, in the case that we don't have a thread_id.

Issue #2424, opened 2025-08-01
Add 'voice instructions' field to VI


fwild:
Allow to enter `instructions` for the TTS voice generation call,
for example to enter "Speak in a cheerful and positive tone."

See also:
https://github.com/WEKIT-ECS/lxr/issues/139

fwild:
Test your own: https://www.openai.fm/

fwild:
Voice instructions work now, but have still a small bug: voice instructions are only saved, if you select a voice after you have added the voice instructions. It should be saved when instructions are added. 

And when you reopen the edit panel for voice, the prompt is not prefilled, but empty again.

fwild:
Seems the value is not restored into a linked TMP gameobject:

<img width="768" height="880" alt="Image" src="https://github.com/user-attachments/assets/0c82dee7-7d99-4e21-9dce-3a5cf22cc5d6" />

fwild:
works now on mobile. spatial still missing.

PR #2438, opened 2025-08-28
2437 instructor thread_id and 2424 voice instructions

cawke:
https://github.com/WEKIT-ECS/MIRAGE-XR/issues/2437
https://github.com/WEKIT-ECS/MIRAGE-XR/issues/2424

The instructor initialization now supports an asynchronous mechanism with `InitResponse`, integrating assistant and thread IDs for AI interactions. Refactored AI-related methods to streamline communication and ensure consistency across various instructor operations. Removed unused packages and updated project configurations for better resource management.

https://github.com/WEKIT-ECS/lib-lee/pull/57


Packages/packages-lock.json
```diff
@@ -299,51 +299,6 @@
...
-      "dependencies": {
-        "com.unity.ugui": "2.0.0",
-        "com.unity.collections": "2.4.3",
-        "com.unity.inputsystem": "1.10.0",
-        "com.unity.shadergraph": "17.0.3",
-        "com.unity.textmeshpro": "3.0.7",
-        "com.unity.ext.flatsharp": "1.1.1",
-        "com.unity.modules.video": "1.0.0",
-        "com.unity.xr.core-utils": "2.3.0",
-        "com.unity.nuget.newtonsoft-json": "3.2.1",
```
fwild: Careful - probably better to remove this change to packages-lock.json overall - as we are using the gltfast fix for polyspatial 2.0.4, and as in the past there was a weird downgrade depdency to newtonsoft json (was it for this version or an earlier one?). Anything we break here, will either cause compile errors when compiling for VisionPro, or create broken shaders (again), rendering all gltf 3D objects black.

Packages/manifest.json
```diff
@@ -59,7 +59,6 @@
     "com.unity.multiplayer.center": "1.0.0",
     "com.unity.nuget.mono-cecil": "1.11.5",
     "com.unity.nuget.newtonsoft-json": "3.2.1",
-    "com.unity.polyspatial.visionos": "2.0.4",
```
fwild: see comment below for packages-lock.json

Assets/MirageXR/ContentTypes/VirtualInstructor/VirtualInstructor.cs
```diff
@@ -12,6 +13,8 @@ namespace MirageXR
     /// </summary>
     public class VirtualInstructor : MirageXRPrefab, IVirtualInstructor
     {
+        private static IArtificialIntelligenceManager ArtificialIntelligenceManager => RootObject.Instance.LEE.ArtificialIntelligenceManager;
+
         public event Action<AudioClip> OnInstructorResponseAvailable;
         
         private const float CharacterHeight = 1.8f;
```
fwild, line 20: HistoryFormat below in line 21 is also no longer needed, right?

fwild (approved):
Some small changes needed - I am worried about the package lock update of polyspatial etc.

Packages/packages-lock.json
fwild: to be precise: we use a different gltfast package that requires polyspatial 2.0.4, so in theory the change here should not affect that. But: unless we compiled, we cannot be 100% sure? I tried to google the newtonsoft json dependency, but could not find any info in the package description or changelog (also for the visionos package). Maybe this was resolved by now?

Assets/MirageXR/ContentTypes/VirtualInstructor/VirtualInstructor.cs
cawke: yes, fixed

Packages/packages-lock.json
cawke: my mistake, fixed

Packages/manifest.json
cawke: my mistake, fixed

fwild:
please update pointer to lib-lee develop (following merge of your PR)

PR #2452, opened 2025-09-12
vision os: on immersion changed event fix

cawke:


Assets/MirageXR/Managers/RoomTwinManager/RoomTwinManager.cs
```diff
@@ -29,128 +26,125 @@ public enum RoomTwinStyle:ushort
...
-            await LoadRoomTwinModel(Path.Combine(Application.streamingAssetsPath, RoomFile));
+            await LoadRoomTwinModelAsync(Path.Combine(Application.streamingAssetsPath, RoomFile));
 
             UnityEngine.Debug.Log("[RoomTwinManager] registering ImmersionChanged event");
-            #if UNITY_VISIONOS || VISION_OS
-                var VolumeCamera = GameObject.Find("/Start").GetComponent<VolumeCamera>();
-                if (VolumeCamera != null) VolumeCamera.ImmersionChanged += OnImmersionChanged;
-            #endif
-
+//#if UNITY_VISIONOS || VISION_OS
```
fwild: is this ok without the pragma #if? wouldn't that interfere on other platforms?

fwild (approved):

Assets/MirageXR/Managers/RoomTwinManager/RoomTwinManager.cs
cawke: fixed

PR #2468, opened 2025-09-16
2460 audio editor and homescreen prefabs had missing scripts

### LINKED ISSUE #2460: AudioEditor_Spatial prefab has missing behaviour script
and HomeScreen_Spatial prefab is missing another: The referenced script on this Behaviour (Game Object 'RecordButtonVITep') is missing!

fwild:
- fixed AudioEditor and HomeScreen prefabs, as they had missing scripts. Seems to all work, but needs careful review? How did it happen?

resolves #2460


fwild:
Hm. I see a lot of changed font thingies - and I did not commit the changed font files, as they should not need to be changed. So: not quite sure if this PR is correct. What I did is remove all broken components from the gameobjects in both prefabs - I don't think they were meant to be there, it seemed like a systemic faulty addition.

cawke (approved):

Issue #2453, opened 2025-09-12
iOS: artefacts on shadows/lighting


fwild:
In the past few weeks, the builds on iOS always have rendering artefacts:

<img width="2388" height="1668" alt="Image" src="https://github.com/user-attachments/assets/a3c11994-645c-4ff1-a946-92fa5c0f6a50" />

Seems some quality settings are not ok? Or is it the gltf importer fix for visionOS interfering?

Issue #2427, opened 2025-08-09
iOS build: rendering problem


fwild:
3D models and virtual instructor render oddly with flickering mosaic pixelations. These errors appear in the log:

```
ExceptionManager: [Error] RenderPass: Attachment 0 was created with 4 samples but 1 samples were requested., trace: UnityEngine.Rendering.RenderPipelineManager:DoRenderLoop_Internal(RenderPipelineAsset, IntPtr, Object)
```
https://wekit-ecs.sentry.io/issues/6796226006/events/651bb097a3024f2dbd66025b90d3fcc2/?project=4506320009428992

```
ExceptionManager: [Error] EndRenderPass: Not inside a Renderpass, trace: UnityEngine.Rendering.RenderPipelineManager:DoRenderLoop_Internal(RenderPipelineAsset, IntPtr, Object)
```
https://wekit-ecs.sentry.io/issues/6547291221/events/f75a1ad2ba914c5494a23a69838a9c50/?project=4506320009428992

PR #2489, opened 2025-11-10
shadow jitter fix

### LINKED ISSUE #2488: shadows jittering pixelated
In URP, changing the rendering path seems to fix it.
Image

It may also be worth setting this additionally: 
Change on the "directional lighting" GameObject in `Start.scene` the shadow type to 'soft shadows':

Image

fwild:
* URP asset renderer: rendering path now set to deferred (was forward+)
* direction light: shadow type = soft shadow (was hard shadow)

resolves #2488 and #2453 and #2427


fwild:
obsolete - seems it was fixed by update to newer unity version (and packages)

PR #2498, opened 2025-12-16
update to fusion 2.0.9 and voice 2.63, downgrade for psp 2.0.4

fwild:
resolves https://github.com/WEKIT-ECS/MIRAGE-XR/issues/2499 and https://github.com/WEKIT-ECS/MIRAGE-XR/issues/2495


fwild:
still crashes when starting collaboration, everything else seems ok. Mic permission is only requested when adding a audio augmentation, and it might be required to select an avatar under profile first - but none of these made a difference, still crashes when starting collab.

PR #2509, opened 2026-02-06
VI recording limit increase (mobile)

### LINKED ISSUE #2421: VI: increase VI input limit to 60 secs (currently: 20)
On VisionPro, after 10 seconds or so, the button changes to stop, writing overlapping text on the button. The subsequent call to the stt endpoint fails with an internal server error, and no response is played to the user.

This seems not to be the case on mobile.

fwild:
- Increased the limit for the recording in the record button inspector setting in the View prefab from 25 to 60 secs

resolves #2421


Assets/MirageXR/Design/MobileUI/Prefabs/View.prefab
```diff
@@ -182,1653 +182,2092 @@ PrefabInstance:
...
+    - target: {fileID: 8449454361001110736, guid: 386c0a96890c32b4e82c463598ce4715,
+        type: 3}
+      propertyPath: 'm_ActiveFontFeatures.Array.data[0]'
       value: 1801810542
       objectReference: {fileID: 0}
-    - target: {fileID: 8471274404863053440, guid: 386c0a96890c32b4e82c463598ce4715, type: 3}
+    - target: {fileID: 8450635616900776560, guid: 386c0a96890c32b4e82c463598ce4715,
+        type: 3}
+      propertyPath: _maxRecordTime
+      value: 60
```
fwild, line 2181: this is the only change I made - the rest are automated changes.

Issue #2472, opened 2025-09-17
Spatial hyperlinks only displayed when task card folded out


fwild:
The spatial hyperlinks are only initialised the second time around, when the task card
is folded out. When collapsed again, they will correctly display (with the new centre),
but when moving steps, the number of spatial hyperlinks is the number from the previous step.

So: small bug, needs proper checking and initalisation.

PR #2512, opened 2026-02-12
Work in progress: spatial hyperlinks bugfix

fwild:
- fixed the bug why when changing steps, the spatial hyperlinks would not be updated properly
- but: found that the logic separating editing (in the controller) and view does not fully work, thus requiring a dirty direct FindObject update call that still needs resolving.

Partially resolves: #2472

### DIFF
--- a/Assets/MirageXR/Player/Scripts/Spatial/StepsScreenSpatialViewController.cs
+++ b/Assets/MirageXR/Player/Scripts/Spatial/StepsScreenSpatialViewController.cs
@@ -212,6 +212,7 @@ namespace MirageXR
             UpdateHyperlinkPrefabs(); 
             await UpdateInfoMediaViewAsync();
             UpdateInfoToolsView();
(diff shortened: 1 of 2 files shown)


Assets/MirageXR/Player/Scripts/Spatial/StepsScreenSpatialViewController.cs
```diff
@@ -212,6 +212,7 @@ private async UniTask UpdateInfoViewAsync()
             UpdateHyperlinkPrefabs(); 
             await UpdateInfoMediaViewAsync();
             UpdateInfoToolsView();
+            FindObjectOfType<InfoScreenSpatialView>()?.UpdateSplines(); // not happy with this
```
fwild, line 215: Not happy with this - I tried moving the whole logic of creating the hyperlink anchor prefabs (the spheres) into the InfoScreenSpatialView, but this did not fully work.

fwild (commented):
Comment added

Issue #2529, opened 2026-04-23
delete step


fwild:
add new icon in the step list to delete a step.

Issue #2568, opened 2026-05-29
Implement safety confirmation for delete


fwild:
Trash icon is dangerous; needs additional confirmation step or hold-to-delete pattern

PR #2584, opened 2026-06-25
improve delete buttons for activity and step

Herman1412:
Activity delete button requires the user to hold it down for a short period before it will trigger the delete. 
Added delete button to step list and gave it the same behavior. (The OnButtonMenuClick also deletes the item, I assume it is supposed to do something else so marked it as TODO)

I have made some changes to ButtonLongPress I could not find any references to it so I assumed it was not being used.

Close: #2529 and #2568


fwild:
Maybe add the confirm dialogue? And is it possible to have the long-press on the whole entry, and remove the trash icon?

fwild:
Activity list: long press; step list: ... menu

fwild:
I think it would be better to have the step delete the same way as the activity list long press - for consistency reasons. Also: the three bar icon is reserved for moving steps (I almost did not find where the context menu is attached. Can we do that, @fominykh? The long press (with the warning dialogue) works really nicely, and is super smooth! Step delete should also have such confirmation dialogue - it is a non-reversible operation, too!

fwild (approved):

